PR #15, opened 2022-11-09
[cosim] check vrf write after commit

SharzyL:
Because rtl may continue write vrf even after committed, it is
necessary to delay the vrf-all-write check a while after commit.


SharzyL:
Decided to detect vrf write on write queue, abandon this plan.

PR #22, opened 2022-11-10
Address calculation for seg-load.

qinjun-li:


sequencer (approved):
timing here is tight....

PR #39, opened 2022-11-20
[cosim] Make simulation exitable

ksco:


tests/emulator/src/vbridge_impl.cc
```diff
@@ -206,6 +227,11 @@ void VBridgeImpl::run() {
         throw TimeoutException();
       }
     }
+
+    if (exit_after_all_issued) {
+      LOG(INFO) << fmt::format("[{}] all insn in to_rtl_queue are issued, exiting", get_t());
+      exit(0);
```
SharzyL, line 233: Avoid `exit`, throw exception, or return to exit loop, in order to ensure graceful cleanup

tests/emulator/src/vbridge_impl.cc
```diff
@@ -155,11 +169,18 @@ void VBridgeImpl::run() {
...
+    bool exit_after_all_issued = false;
     // loop while there exists unissued insn in queue
     while (!to_rtl_queue.front().is_issued) {
       // in the RTL thread, for each RTL cycle, valid signals should be checked, generate events, let testbench be able
       // to check the correctness of RTL behavior, benchmark performance signals.
       SpikeEvent *se_to_issue = find_se_to_issue();
+      if (se_to_issue->is_exit_insn) {
+        se_to_issue->is_issued = true;
+        exit_after_all_issued = true;
+        continue;
```
SharzyL, line 181: Maybe we should prevent instructions being issued after exit instruction?

ksco: The exit instruction is guaranteed to be the last one in the queue.

tests/emulator/src/vbridge_impl.cc
ksco: Ack

tests/emulator/src/vbridge_impl.cc
```diff
@@ -128,12 +134,20 @@ std::optional<SpikeEvent> VBridgeImpl::create_spike_event(insn_fetch_t fetch) {
...
+
   // for load/store instr, the opcode is shared with fp load/store. They can be only distinguished by func3 (i.e. width)
   // the func3 values for vector load/store are 000, 101, 110, 111, we can filter them out by ((width - 1) & 0b100)
   bool is_load_type  = opcode == 0b0000111 && ((width - 1) & 0b100);
   bool is_store_type = opcode == 0b0100111 && ((width - 1) & 0b100);
   bool v_type = opcode == 0b1010111 && width != 0b111;
-  if (is_load_type || is_store_type || v_type) {
+
+  bool csr_type = opcode == 0b1110011 && (width & 0b011);
+  bool csr_write = csr_type && ((width & 0b100) | rs1);
```
SharzyL, line 147: I prefer using `is_xxx` variable name for boolean

ksco: Ack

SharzyL (commented):
A few changes suggested

tests/emulator/src/vbridge_impl.cc
ksco: Agreed. Is it necessary to log the exit message? If it's not, we can return from here.

ksco: Agreed. Is it necessary to log the exit message? If it's not, we can return from here.

SharzyL: edited

SharzyL: log then exit seems fine

SharzyL (approved):

PR #73, opened 2022-12-26
[cosim] add sigint handler

SharzyL:


ksco (approved):

tests/emulator/src/dpi.cc
```diff
@@ -33,6 +40,7 @@ void VBridgeImpl::dpiDumpWave() {
 }
 
 [[maybe_unused]] void dpiInitCosim() {
+  std::signal(SIGINT, sigint_handler);
```
ksco, line 43: According to https://stackoverflow.com/a/1641223, `sigaction` seems to be a more reliable way.

SharzyL: `std::signal` is an ISO C++ standard, while `sigaction` is POSIX standard. Hard to say which is more reliable.

PR #86, opened 2023-01-12
vaadd test

qinjun-li:
- [rtl] Handling sign bits read from scalar.
- [rtl] Handling overflow for average.


sequencer (approved):

qinjun-li:
这个 pr 是在修 average 类型的加法指令

PR #125, opened 2023-02-23
[build system] add buddy-mlir to nix

SharzyL:


sequencer (approved):
That’s an astonishing cp

PR #140, opened 2023-02-28
Buddy mlir integration

ksco:


build.sc
```diff
@@ -416,16 +416,80 @@ object tests extends Module {
     object smoketest extends Case
 
     object mmm extends Case
+
+    object buddy extends Case {
```
sequencer, line 420:
```suggestion
    object buddy extends Cross[BuddyMLIRCase]("hello.mlir") 
    class BuddyMLIRCase(mlirSourceFile: String) extends Case {
```

sequencer (changes requested):
Please detect case based on file name

sequencer (dismissed):

qinjun-li (approved):

PR #145, opened 2023-03-02
multiplyAdd

qinjun-li:
- [rtl] Handling data types of ma.
- [rtl] Handling ma-type instructions for reading vd across lane.


sequencer (approved):
todo: extract datapath from when

PR #149, opened 2023-03-05
[perf] add alu/lsu/vrf_write perf

SharzyL:


tests/elaborate/src/Monitor.scala
```diff
@@ -0,0 +1,122 @@
...
+    peeker.valid := tap(lane.vrf.write.valid)
+    peeker.clock := clock
+  }
+
+  // alu occupied
+  dut.laneVec.zipWithIndex.foreach { case (lane, index) =>
+    val peeker = Module(new ExtModule with HasExtModuleInline {
+      override val desiredName = "dpiALUMonitor"
+      val clock = IO(Input(Clock()))
+      val landIdx = IO(Input(UInt(32.W)))
```
sequencer, line 46: lane

ksco (approved):
Maybe we can calculate an overall score?

PR #168, opened 2023-03-15
[cosim] merge spike mem write

ksco:


sequencer:
We need merging read as well.

PR #207, opened 2023-04-12
Add custom asm test case and pipeline

Origami404:
Just copied from MLIR test pipeline. 

Use `.asm` as test case identify extension name . 

Can be used as a temporary measure until C RVV intrinsic is supported.


Origami404:
As #216 is merged, this PR should be closed.

PR #210, opened 2023-04-18
Add test cases for rvv-vp-intrinsic-add and rvv-vp-intrinsic-add-scal…

xlinsist:
…able

# Summary
**For now there are errors to be resolved** when running either rvv-vp-intrinsic-add-scalable.mlir or rvv-vp-intrinsic-add-scalable.mlir.


sequencer:
Will be fixed in #212

PR #229, opened 2023-05-25
[buddy] Use llvm-lit like test case runner

Avimitin:
Different test cases require different lowering passes, and the original implementation hard-coded all passes for all cases, which limited extensibility.

This commit adds a comment parser for mill to parse the header comment of each mlir test case to get case-specific buddy-opt argument. After this commit, developers can place passes into a block starting with keyword "BUDDY-OPT" and ending with keyword "BUDDY-OPT-END".

Here is an example for passing "--lower-rvv=rv32" to buddy-opt.

```mlir
// BUDDY-OPT
// --lower-rvv=rv32
// BUDDY-OPT-END

// ... mlir code ...
```

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -404,8 +404,8 @@ object tests extends Module {
       // These arguments must be wrapped in a block, starting with "BUDDY-OPT" and ending with "BUDDY-OPT-END".
       def parseBuddyOptArg(testFile: os.Path) = os.read
             .lines(testFile)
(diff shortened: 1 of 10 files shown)


sequencer (approved):
LGTM

Avimitin:
Please don't merge this now, I think the block boundary checker can be improve. I will push another commit soon and you can squash them together.

sequencer (approved):

PR #235, opened 2023-06-14
[buddy] remove redundant case

Avimitin:
The stripmining.mlir case is an old handwritten case as a workaround for the lack of the lower-rvv pass. But now we have stripmining-huge.mlir, which serves the same purpose but has better implementation than the former case, so we don't need the stripminig.mlir case anymore.


xlinsist:
LGTM

SharzyL (approved):

PR #239, opened 2023-06-19
[rtl] Merge continuously accessed tile-link requests.

qinjun-li:


sequencer (approved):
One giant leap for RISC-V Vector

PR #243, opened 2023-06-21
intensive chaining

qinjun-li:
- feat[cosim]: vsetvl no longer blocks the pipeline
- [rtl] Handle WAR when st followed by the normal instruction.
- [rtl] The instruction needs to use its own VSEW to regroup v0.


sequencer (approved):
performance!

PR #247, opened 2023-07-03
[build system] merge impl and elaborate together

sequencer:
This PR move configuration to build time, to be read by both RTL generator and emulator


sequencer:
Merge this for now.

PR #251, opened 2023-07-07
[build system] bump circt to 1.44.0

sequencer:


sequencer:
It should take a long time to build.

PR #264, opened 2023-07-21
[test] Enable RISC-V F in assembly to support vector FP operations

xlinsist:
When testing mlir demos in vector repo, the generated assembly will be called in file "main.s", where defines which RISC-V extended instruction set to use. Though FP operations has been supported in hardware's side, RISC-V "F" hasn't been enabled for testing:< which leads to `spike illinstr` when running FP vector operations in buddy cases. This PR enables RISC-V "F" in assembly for all the tests.


sequencer:
Glad you successfully get it done :)

PR #269, opened 2023-07-27
[test] Replace RVV intrinsics with built-in functions

xlinsist:
Replace the RVV intrinsics like `vsetvl_e32m4` in matmul.c with something like `__riscv_vsetvl_e32m4` to ensure that the program can be compiled normally.  `__riscv_vsetvl_e32m4` is one of the built-in functions provided by the RISC-V LLVM backend.


Avimitin:
This PR pass successfully in my local machine and it blocks the testplan branch, so it is ok to be merged without passing the test. (Also the fail tests are failed because there is no space left on the CI machine.)

Avimitin (approved):

PR #272, opened 2023-07-27
[intrinsic] fix matmul test for latest LLVM

Origami404:
fix like the bf0cb8f42468736774c0e55e92d08c720072ac63


Avimitin:
This seems to conflict with #269.

PR #278, opened 2023-08-01
[ci] allow user select verilator and run config

Avimitin:
Merge #276 before this.


Makefile
```diff
@@ -40,7 +40,7 @@ ci-run:
 
 ci-passed-tests:
 	echo -n matrix= >> $$GITHUB_OUTPUT
-	amm .github/scripts/ci.sc passedJson $(RUNNERS) .github/passed.txt ./passed.json
+	amm .github/scripts/ci.sc passedJson $(RUNNERS) $(PASSEDFILE) ./passed.json
```
sequencer, line 43: This idea is good, but consuming more than one file and scheduling them is better.

PR #305, opened 2023-09-02
[cosim] refactor elaborator using DPIModule

SharzyL:


emulator/src/dpi.cc
```diff
@@ -95,8 +97,8 @@ void VBridgeImpl::dpiDumpWave() {
   })
 }
 
-[[maybe_unused]] void dpiPeekTL(
-    int channel_id,
+[[maybe_unused]] void peek_t_l(
```
sequencer, line 100: fk

sequencer: change to peekTl

sequencer (approved):

PR #306, opened 2023-09-02
Run smoke test for subsystem emulator

cyyself:


sequencer (commented):
Please use plusargs to consume the binary path

cyyself:
> Please use plusargs to consume the binary path

OK

PR #315, opened 2023-09-11
Add VectorAdder and VectorMultiplier for banked calculation

midnighter95:


sequencer:
@qinjun-li need to integrate this into t1

PR #327, opened 2023-10-20
[tests] Add newlib toolchain

Avimitin:
Current status:

- [x] sbrk (malloc is verified working correctly)
- [x] exit (verified)
- [x] write
- [x] close
- [x] lseek


Avimitin:
Closed. I want to split the library and DL Model into two different PR.

PR #329, opened 2023-10-24
[build system] build test case when TEST_CASE_DIR is unset

Avimitin:
This commit remove the caseBuild build task and let the verilatorEmulator automatically build test case when the test binary is not found.


Avimitin:
Update:

* remove all build dependencies from .#testcase devShell
* use same testcase finding logic for verilator and subsystememu

Avimitin:
@sequencer This PR is ready to be merged.

PR #332, opened 2023-10-25
[build system] allow using pre-built emulator to run tests

Avimitin:
This PR achieve these goals:

- Add a nix derivation for verilator emulator, which can be cached by nix store and be reused across multiple runners.
- Bump mill to 0.11.5 so that we can pre-fetch dependencies before build, and now all the sandbox option can be drop.
- Update CI logic to generate wave file after test fail, so that our RTL dev can download and debug those failing tests.


sequencer (dismissed):

Avimitin:
This PR is now ready to be merged.

PR #333, opened 2023-10-26
[emulator] improve emulator logging performance

Avimitin:
- allow disable logging functionality
- display cycle data for all log entry
- use global logger to avoid construct/deconstruct thousands of logger instance


Avimitin:
I want to merge this into #334 since these two PR are related.

PR #337, opened 2023-10-31
performance

qinjun-li:
- [rtl] Remove unnecessary reads.
- [rtl] Detect load unit write-after-write conflicts in advance.
- [test] Change smoke.asm length to test the performance of accessing vrf.

### DIFF
--- a/tests/asm/smoke.asm
+++ b/tests/asm/smoke.asm
@@ -1,23 +1,31 @@
 .global test
 test:
     addi x1, x1, 1
-    vsetivli x0, 16, e8, m1, ta, ma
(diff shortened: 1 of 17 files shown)


qinjun-li:
1. Let multiple source operands read the vrf pipeline separately.
2. LSU does WAW detection before writing vrf.
3. Update the record of load instruction in vrf more accurately(The segment type only updates the first group)
4. Separate the operation of splitting data group, let non-combinable execution units split themselves.
5. Let the execution stage also be pipelined.

sequencer (approved):

PR #341, opened 2023-11-07
Add UART to Emulator

cyyself:


emulator/src/simple_sim.h
```diff
@@ -26,17 +29,31 @@ class simple_sim : public simif_t {
...
+      return NULL;
+    }
     CHECK_LE(addr, mem_size,
              fmt::format("memory out of bound ({:016X} >= {:016X})", addr,
                          mem_size));
     return &mem[addr];
   }
 
   bool mmio_load(reg_t addr, size_t len, uint8_t *bytes) override {
+    if (uart_addr <= addr && addr <= uart_addr + 16) {
```
ksco, line 42: Might be better to use `sizeof(uartlite_regs)` than `16`.

emulator/src/simple_sim.h
```diff
@@ -26,17 +29,31 @@ class simple_sim : public simif_t {
...
+      return uart.do_read(addr - uart_addr, len, bytes);
+    }
     FATAL("Unimplemented");
   }
 
   bool mmio_store(reg_t addr, size_t len, const uint8_t *bytes) override {
+    if (uart_addr <= addr && addr <= uart_addr + 16) {
```
ksco, line 49: Might be better to use `sizeof(uartlite_regs)` than `16`.

emulator/src/uartlite.h
```diff
@@ -0,0 +1,93 @@
...
+    unsigned int rx_fifo;
+    unsigned int tx_fifo;
+    unsigned int status;
+    unsigned int control;
+};
+
+class uartlite : public mmio_dev  {
+public:
+    uartlite() {
+        memset(&regs,0,sizeof(regs));
```
ksco, line 27: nit: Add spaces after `,`, same below.

emulator/src/uartlite.h
```diff
@@ -0,0 +1,93 @@
...
+    }
+    bool do_read(uint64_t start_addr, uint64_t size, unsigned char* buffer) {
+        std::unique_lock<std::mutex> lock(rx_lock);
+        if (start_addr + size > sizeof(regs)) return false;
+        if (!rx.empty()) {
+            regs.status |= SR_RX_FIFO_VALID_DATA;
+            regs.rx_fifo = rx.front();
+        }
+        else regs.status &= ~SR_RX_FIFO_VALID_DATA;
+        memcpy(buffer,((char*)(&regs))+start_addr,std::min(size,sizeof(regs)-start_addr));
```
ksco, line 39: `size` is already checked at line 33, the `std::min` here seems redundant, we can use size directly. Same on `do_write`.

emulator/src/uartlite.h
```diff
@@ -0,0 +1,93 @@
...
+            if (regs.control & ULITE_CONTROL_RST_TX) {
+                while (!tx.empty()) tx.pop();
+            }
+            if (regs.control & ULITE_CONTROL_RST_RX) {
+                while (!rx.empty()) rx.pop();
+            }
+        }
+        return true;
+    }
+    void putc(char c) {
```
ksco, line 64: Unused.

emulator/src/uartlite.h
```diff
@@ -0,0 +1,93 @@
...
+            if (tx.empty()) wait_ack = true;
+            return res;
+        }
+        else return -1;
+    }
+    bool exist_tx() {
+        std::unique_lock<std::mutex> lock(tx_lock);
+        return !tx.empty();
+    }
+    bool irq() {
```
ksco, line 82: Unused.

tests/intrinsic/uart.h
```diff
@@ -0,0 +1,30 @@
...
+#include <stdint.h>
+
+struct uartlite_regs {
+    volatile unsigned int rx_fifo;
+    volatile unsigned int tx_fifo;
+    volatile unsigned int status;
+    volatile unsigned int control;
+};
+
+struct uartlite_regs *const ttyUL0 = (struct uartlite_regs *)0x60000000;
```
ksco, line 12: `ttyUL0` should be more appropriate to be volatile than `struct uartlite_regs` members.

tests/intrinsic/uart.h
```diff
@@ -0,0 +1,30 @@
+#pragma once
+
+#include <stdint.h>
+
+struct uartlite_regs {
+    volatile unsigned int rx_fifo;
+    volatile unsigned int tx_fifo;
+    volatile unsigned int status;
+    volatile unsigned int control;
+};
```
ksco, line 10: Can we reuse `struct uartlite_regs` and the constants with `emulator/src/uartlite.h`?

emulator/src/uartlite.h
```diff
@@ -0,0 +1,93 @@
+#pragma once
+
+#include "mmio_dev.h"
+#include <cstring>
+#include <algorithm>
+#include <queue>
+#include <mutex>
+
+#define SR_TX_FIFO_FULL         (1<<3) /* transmit FIFO full */
```
ksco, line 9: Looks like our simple uart implementation is never full on rx or tx, add a comment to note here.

emulator/src/uartlite.h
```diff
@@ -0,0 +1,93 @@
...
+        std::unique_lock<std::mutex> lock(rx_lock);
+        if (start_addr + size > sizeof(regs)) return false;
+        if (!rx.empty()) {
+            regs.status |= SR_RX_FIFO_VALID_DATA;
+            regs.rx_fifo = rx.front();
+        }
+        else regs.status &= ~SR_RX_FIFO_VALID_DATA;
+        memcpy(buffer,((char*)(&regs))+start_addr,std::min(size,sizeof(regs)-start_addr));
+        wait_ack = false;
+        if (start_addr <= offsetof(uartlite_regs,rx_fifo) && offsetof(uartlite_regs,rx_fifo) <= start_addr + size) {
```
ksco, line 41: This pattern appears 3 times, can we make it a macro or something?

emulator/src/uartlite.h
cyyself: I will reserve it for future UART input function needs.

emulator/src/uartlite.h
cyyself: Same as getc function.

tests/intrinsic/uart.h
cyyself: I will preserve this due to the uartlite driver is common in most system software including opensbi, u-boot, linux.

ksco (approved):

PR #340, opened 2023-11-06
[build system] refactor nix sources, split emulator build to nix

SharzyL:
## Checklist
- [x] generate elaborator with nix
- [x] generate emulator with nix
- [ ] ~~compile testcases without mill~~
- [ ] ~~add documentation for testcase memory map~~
- [x] remove git submodule, populate submodules on shellhook
- [x] setup emulator runner script
- [x] let CI work again
- [x] add doc of nix interface

PR #347, opened 2023-11-13
[nix] allow verilator emulator always use latest source

Avimitin:
This PR force the verilator-emulator derivation use the latest RTL sources, so that we can verify the correctness of the latest RTL.


Avimitin:
Closed since #340 is merged.

PR #350, opened 2023-11-14
[tests] add memcpy-asm rvv test case

vowstar:
Added relevant test cases for memcpy. Corresponding config files and asm files have been added, reusing main.S. Since the current CI only tests for RTL bugs and does not check the correctness of the case itself, this case will execute to the end regardless of the result. However, for future correctness testing, I have added a result at the end of the test that will be convenient for future maintainers to handle:

- if the test passes, the `a0` register will be set to `0xff`
- if the test fails, the `a0` register will be set to `0xfe`

future maintainers can continue to add assembly code to decide specifically what to do in case of test pass or fail.

### DIFF
--- a/tests/asm/memcpy.asm
+++ b/tests/asm/memcpy.asm
@@ -1,0 +1,55 @@
+# src: https://github.com/riscv/riscv-v-spec/blob/master/example/memcpy.s
+# modifications:
+# 1. add test to call memcpy
+# 2. add memset to fill memory
+# 3. add exit to finish simulation
+
+.text
+.balign 16
+.globl test
+
+# void *memcpy(void* dest, const void* src, size_t n)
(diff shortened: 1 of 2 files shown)


sequencer:
Basically we don't need to add check since we are using difftest to check between spike and rtl

vowstar:
So that's it, looks like I can remove the code that checks the memory.

sequencer:
Yes, just add pure test. Our test framework can make sure if rtl is correct
or not.

vowstar:
Done!

sequencer:
nice, @qinjun-li to make sure chaining will work in this case.

sequencer (approved):

sequencer:
free to skip CI

PR #352, opened 2023-11-15
[ci] use last release to compare test dir changes

Avimitin:


Avimitin:
Closed since #340 is merged.

PR #355, opened 2023-11-16
Refactor and fix the build system changes

Avimitin:
- [x] Bump nixpkgs to latest rev(e44462d6021bfe23dfb24b775cc7c390844f773d) for latest mill and jdk21
- [x] remove mill runtime derivation since the rt.jar file is not reproducible
  * rt.jar contains timestamp and machine specific info
  * rt.jar have some machine specific bytecode generate in internal system module.

<br>

---
- [x] remove all `leaveDotGit` option in dependencies: https://github.com/NixOS/nixpkgs/issues/8567
- [x] finish the changes to the release script
- [x] fix invalid file order generated by firtool
- [x] fix latest fmt lib doesn't support formatting json value error
- [x] remove wal due to its bad build system


.github/scripts/ci.sc
```diff
@@ -219,7 +219,7 @@ def runTests(jobs: String, resultDir: Option[os.Path]) = {
       System.err.println(s"[${i+1}/${totalJobs.length}] Running test case $config,$caseName,$runCfg")
       val handle = os
         .proc("scripts/run-test.py", "-c", config, "-r", runCfg, "--no-console-log", "--base-out-dir", testRunDir, caseName)
-        .call(check=false)
+        .call(check=false, stdout=os.Path("/dev/null"))
```
SharzyL, line 222: Is there anything printed to stdout?

nix/t1/submodules.nix
```diff
@@ -15,27 +15,28 @@ in
...
 
-      if ! git -C "$depDir" remote get-url origin >/dev/null 2>&1; then
-        echo "[nix-shell] adding remote '$remote' for '$depDir'"
-        git -C "$depDir" remote add origin "$remote"
+        git clone $remote $depDir
+        pushd $depDir > /dev/null
+        git -c advice.detachedHead=false checkout $rev
+        popd > /dev/null
+
+        echo "[nix-shell] checkout to *detached head*: $rev for submodule $name"
```
SharzyL, line 38: Logging here seems unnecessary here since git checkout would print the commit being checked out

SharzyL (commented):
Just some nitpicks

.github/scripts/ci.sc
Avimitin: Those Python script logs that make it hard to read which test failed.

SharzyL: Well, but log goes to stderr

PR #373, opened 2023-12-12
[build system] remove prebuilt

SharzyL:


sequencer:
aarch64 doesn't work for now, need to fix M1/M2 compiling.

PR #385, opened 2023-12-20
soc memcpy

qinjun-li:
- [soc] fix setvli bug.
- [soc] vop.v.x need read rs1.
- [nix] add mill deps hash for soc.


sequencer:
why performance changed...

qinjun-li:
In the past, there was back pressure on the write queue?

PR #392, opened 2023-12-29
[rtl] Directly connected cross-lane reading and writing.

qinjun-li:


sequencer (approved):
Ultimate way to speed up cross lane instructions

PR #400, opened 2024-01-03
[rtl] Slots can always be shifted when slot idle.

qinjun-li:


qinjun-li:
Continue to complete this part in 602

PR #401, opened 2024-01-03
[ci] add implementation for transforming test results as derivation

Avimitin:
After this PR, users can get the test outputs (perf data, wave.fst...etc) by using command `nix build .#t1.<config>.<emulator>.<case-name>`


Avimitin:
Add some new modification: I think the colon ":" between case type and case name is really confusing. So I decided not to flatten the test case structure. I think `nix build .#t1.<config>.ip/subsystem.emu.<caseType>.<caseName>` might be better to understand.

For example, if user wants to get perf result and wave file of memcpy-asm on ip emulator with v1024 l8 b2 configured, they can run `nix build .#t1.v1024-l8-b2.ip.emu-trace.asm.memcpy`.

This might seem too long at first glance, but since we have `nix search` support now, it is not hard to get what is runnable for all kind of emulator.

PR #402, opened 2024-01-03
[nix] add search ability for derivation in our t1 scope

Avimitin:
Set `recurseForDerivation = true` allow nix search to search inside attr set recursively. Now users can use `nix search .#t1 <regexp>` to search through all of the derivations we make.


sequencer (approved):

sequencer:
Magic...

PR #403, opened 2024-01-04
[cosim] add soc cosim framework

midnighter95:
Add soc cosim framework which compares  all the scala rf writing with Spike.


sequencer:
we decide to go to offline diff test.

PR #319, opened 2023-09-13
[chisel] Add support for Chisel Binder framework

SpriteOvO:

PR #404, opened 2024-01-04
Use chisel binder to emit Verilog

sequencer:
- [build system] track chisel ci nightly for now
- [build system] bump dependency
- [build system] setup binder for elaborator
- [nix] setup nix for binder
- [nix] use binder to elaborate rtl by default

### DIFF
--- a/nix/t1/elaborate.nix
+++ b/nix/t1/elaborate.nix
@@ -10,6 +10,7 @@
 , elaborator
 , config-name
 , target
+, use-binder
 }:
 
 assert lib.assertMsg
@@ -28,8 +29,6 @@ let
     "--ip-config"
     # Can't use `toString` here, or due to some shell escape issue, Java nio cannot find the path
     "${elaborate-config}"
-    "--target-dir"
-    "elaborate"
   ] ++ lib.optionals (lib.elem target [ "subsystem" "subsystememu" "fpga" ]) [ "--rvopcodes-path" "${riscv-opcodes-src}" ];
 in
 stdenvNoCC.mkDerivation {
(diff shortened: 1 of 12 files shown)


sequencer:
this will take over #319

nix/t1/elaborate.nix
```diff
@@ -28,30 +29,31 @@ let
...
   name = "t1-${config-name}-elaborate-${target}";
   nativeBuildInputs = [ jq espresso circt ];
   buildCommand = ''
     mkdir -p elaborate $out
-    ${elaborator}/bin/elaborator ${target} ${lib.escapeShellArgs elaborateArgs}
+    ${elaborator}/bin/elaborator ${target} ${lib.escapeShellArgs elaborateArgs} "--target-dir" ${if use-binder then "$out" else "elaborate"} ${lib.optionalString (use-binder) "--use-binder"}
 
-    # TODO: use binder to replace emiting below
-    firtool elaborate/*.fir --annotation-file elaborate/*.anno.json -o $out ${mfcArgs}
+    if [ ${lib.boolToString use-binder} = false]; then
```
Avimitin, line 41: Move the condition from bash to nix. Use `lib.optionalString` to produce empty string when use-binder = false.

nix/t1/elaborate.nix
```diff
@@ -28,30 +29,31 @@ let
...
-    #
-    # Fix file ordering difference introduced in some unknown breaking change between firtool 1.50 -> 1.58
-    # In the previous working version, all files starting with './' should be placed on top of the filelist.f.
-    # However in the latest version, they were placed at the bottom, which breaks the verilator.
-    # Here is an disgusting workaround to make it work. But we need to fix this issue at firtool side.
-    echo "Fixing generated filelist.f"
-    grep '^\./' $out/filelist.f > prefixed.f
-    grep -v '^\./' $out/filelist.f > not-prefixed.f
-    cat prefixed.f not-prefixed.f > $out/filelist.f
+    if [ ${lib.boolToString use-binder} = false]; then
```
Avimitin, line 45: Same as above, move this condition to line 44 and create an AND condition.

nix/t1/elaborate.nix
```diff
@@ -10,6 +10,7 @@
 , elaborator
 , config-name
 , target
+, use-binder
```
Avimitin, line 13: Since the argument list is growing, I don't recommend using override in callPackage. It is better to add new argument to this function and use the currying feature to build those derivation.

For example:

```nix
{ pkgA, pkgB, pkgC }:
{ optionA, optionB, optionC}:
derivation
```

```nix
# produce a function attrset -> drv
fn = callPackage ./above_example {}; 
rtl = fn { optionA = true; ...}
```

PR #418, opened 2024-01-20
[ipemu] use cmdline args, cleanup

SharzyL:
- [ipemu] codestyle nitpicks
- [nix] use lib.fileset
- [doc] update emulator develop instruction
- [ipemu] use cmdline instead of env


sequencer:
emmm why cycle changed.

sequencer (approved):

PR #448, opened 2024-02-19
[t1] change double read ports to a read port and a write port

Avimitin:


qinjun-li:
Added config in another way in #450

PR #452, opened 2024-02-21
[nix] bump to firtool 1.68, marking input clock explicit

sequencer:
Bump to nixpkgs master branch since we can depends on our own S3 cache.
This PR will also bump circt to 1.66.0

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -5,11 +5,11 @@
         "systems": "systems"
       },
       "locked": {
-        "lastModified": 1705309234,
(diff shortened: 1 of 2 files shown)


Avimitin:
I think we should keep using the unstable/unstable-small branch, add nixpkgs master branch as another input, and overlay the circt pkg in unstable nixpkgs using this new input.

SpriteOvO (approved):
LGTM

PR #455, opened 2024-02-21
rw port conflict

qinjun-li:
- [rtl] add mem config, pull vrf out
- [rtl] refactor vrf with sram type.
- [rtl] fixup upickle
- [rtl] fix write ready for p0rw.
- [rtl] connect ram port.
- [rtl] fix lastExecuteCounter for mask destination.
- [rtl] fix match bug.
- [rtl] add support for p0rw.
- [rtl] Added conflict checking between reading and writing.


qinjun-li:
Those that can read and write at the same time seem to be blocked by the chaining check.

qinjun-li:
So it seems that just asserting is enough

sequencer (approved):

PR #478, opened 2024-03-16
Support for memory regions

SharzyL:
- **[cosim] overhaul logging infrastructure**
- **[cosim] mask off higher bits on accessing simple_sim**
- **[nix] move emurt to tests/**
- **[cases] let cases use SRAM memory region**
- **[cases] let mlir cases use SRAM mem**


nix/t1/testcases/make-codegen-case.nix
```diff
@@ -20,17 +20,29 @@ stdenv.mkDerivation (rec {
...
 
-    single \
-      -VLEN ${toString vLen} \
-      -XLEN ${toString xLen} \
-      -testfloat3level 2 \
+    ${rvv-codegen}/bin/single \
+      -VLEN "$vLen" \
+      -XLEN "$xLen" \
       -repeat 16 \
+      -testfloat3level 1 \
```
sequencer, line 45: required?

sequencer: I think this was added in master

tests/t1.ld
```diff
@@ -0,0 +1,29 @@
+OUTPUT_ARCH(riscv)
+ENTRY(_start)
+
+MEMORY {
+  SCALAR (RWX) : ORIGIN = 0x20000000, LENGTH = 512M /* put first to set it as default */
+  MMIO   (RW)  : ORIGIN = 0x00000000, LENGTH = 512M
+  DDR    (RW)  : ORIGIN = 0x40000000, LENGTH = 2048M
+  SRAM   (RW)  : ORIGIN = 0xc0000000, LENGTH = 2048M
```
sequencer, line 8: LENGTH cannot be 2048M but anyway it's OK for tests

nix/t1/testcases/make-codegen-case.nix
SharzyL: Seems it should be changed to 2

Avimitin:
This PR is rebase from #480, please merge #480 before this PR.

sequencer (approved):

PR #487, opened 2024-03-23
Provide a unified emulator for both IP and subsystem. 

sequencer:


sequencer:
This will be blocked by the offline diff tests.

PR #495, opened 2024-04-01
Introduce OM to T1 project

SpriteOvO:


omreaderlib/src/Interface.scala
```diff
@@ -0,0 +1,61 @@
...
+  private val cvt = PanamaCIRCTConverter.newWithMlirBc(mlirbc)
+
+  private val pm = cvt.passManager()
+  assert(pm.populatePreprocessTransforms())
+  assert(pm.populateCHIRRTLToLowFIRRTL())
+  assert(pm.populateLowFIRRTLToHW())
+  assert(pm.populateLowHWToSV())
+  assert(pm.populateExportVerilog(_ => ()))
+  assert(pm.populateFinalizeIR())
+  assert(pm.run())
```
sequencer, line 28: I think we don't need to populate to Verilog here?
what we need to do is assume we have got the final mlirbc from firtool binary, evaluate and read from which.
depending on the populate function can be added in the future PR. e.g. remove firtool usage and switch to Scala.

omreaderlib/src/Interface.scala
```diff
@@ -0,0 +1,61 @@
...
+import chisel3.panamaconverter.PanamaCIRCTConverter
+
+object OMReader {
+  def fromFile(mlirbcFile: os.Path, className: String): OMReader = {
+    new OMReader(os.read.bytes(mlirbcFile), className)
+  }
+
+  def fromBytes(mlirbc: Array[Byte], className: String): OMReader = {
+    new OMReader(mlirbc, className)
+  }
```
sequencer, line 15: We don't need `fromBytes` since in our case, we only need to read mlirbc into OMReader.
```suggestion
  private def fromBytes(mlirbc: Array[Byte], className: String): OMReader = {
    new OMReader(mlirbc, className)
  }  
```

omreaderlib/src/Interface.scala
```diff
@@ -0,0 +1,61 @@
...
+    val methods = typeOf[T1Reader].decls.toList.filter(
+        m => m.isPublic && m.isMethod && !m.isConstructor && !m.asMethod.isGetter
+    )
+    methods.foreach(method => {
+      if (!method.name.toString.startsWith("dump")) {
+        val value = mirror.reflectMethod(method.asMethod)()
+        println(s"${method.name} = $value")
+      }
+    })
+  }
```
sequencer, line 56: this is magic... remind all value should be String, and you need to format it on your own.
But do we really need this function?

omreaderlib/src/Interface.scala
```diff
@@ -0,0 +1,61 @@
...
+
+  def dumpAll(): Unit = {
```
sequencer, line 58:
```suggestion
  def dumpOM(): Unit = {
```

sequencer: what's the purpose of this function?

omreader/src/Main.scala
```diff
@@ -0,0 +1,56 @@
...
+    def read(strs: Seq[String]): Either[String, os.Path] = Right(os.Path(strs.head, os.pwd))
+  }
+
+  @main
+  def run(
+    @arg(name = "mlirbc-file") mlirbcFile: Option[os.Path],
+    @arg(name = "class-name") className: String,
+    @arg(name = "dump-methods") dumpMethods: Flag,
+    @arg(name = "eval") eval: Option[String],
+  ) = {
```
sequencer, line 23: I prefer API to be as simple as possible, this API is used by doc/PnR/DV guys. We should statically maintain them.
```
  def dumpOM(
    @arg(name = "mlirbc-file") mlirbcFile: os.Path,
    @arg(name = "class-name") className: String,
  ): String = ???
  def vlen(
    @arg(name = "mlirbc-file") mlirbcFile: os.Path,
    @arg(name = "class-name") className: String,
  ): String = ???
  def dlen(
    @arg(name = "mlirbc-file") mlirbcFile: os.Path,
    @arg(name = "class-name") className: String,
  ): String = ???
```
1. `mlirbcFile`: just use `mlirbcFile`, we generally don't use read from stdin.
2. `className` should be automatically detected, users don't need know what Class is, or force it to be fixed.

omreader/src/Main.scala
```diff
@@ -0,0 +1,56 @@
...
+    input.split("\\.").foldLeft(entry) {
+      case (obj, field) =>
+        if (field.forall(_.isDigit)) {
+          obj.asInstanceOf[PanamaCIRCTOMEvaluatorValueList].getElement(field.toLong)
+        } else {
+          obj.asInstanceOf[PanamaCIRCTOMEvaluatorValueObject].field(field)
+        }
+    }
+  }
+}
```
sequencer, line 56: Don't do it.

sequencer (changes requested):

omreaderlib/src/Interface.scala
SpriteOvO: Populating these passes is essential. Currently T1 only produces mlirbc with `firtool --parse-only`.

omreaderlib/src/Interface.scala
SpriteOvO: This is omreaderlib function, we read mlicbc in cli, then pass the bytes into lib.

omreaderlib/src/Interface.scala
SpriteOvO: This method is for debugging purpose only.

omreaderlib/src/Interface.scala
SpriteOvO: Dump the whole tree of the input instantiated OM object.

omreader/src/Main.scala
SpriteOvO: Makes sense. I have pushed a commit for it, could you review it again please?

omreader/src/Main.scala
SpriteOvO: This is also for debugging purpose only, like `dump-` methods.

omreaderlib/src/Interface.scala
SpriteOvO: I have removed these since now we ran these passes via `firtool` before reading OM.

omreader/src/Main.scala
```diff
@@ -0,0 +1,68 @@
...
+  def vlen(@arg(name = "mlirbc-file") mlirbcFile: os.Path) = {
+    println(simplyGetT1Reader(mlirbcFile).vlen)
+  }
+
+  @main
+  def dlen(@arg(name = "mlirbc-file") mlirbcFile: os.Path) = {
+    println(simplyGetT1Reader(mlirbcFile).dlen)
+  }
+
+  def simplyGetT1Reader(mlirbcFile: os.Path) = OMReader.fromFile(mlirbcFile, "T1Subsystem_Class").t1Reader
```
sequencer, line 52: We need to support both `T1_Class` and `T1Subsystem_Class`

sequencer (changes requested):

nix/overlay.nix
```diff
@@ -14,7 +14,26 @@ in
 
   # Override "nixpkgs" circt with "nixpkgs-for-circt".
   # To update the "nixpkgs-for-circt" input, run `nix flake lock --update-input nixpkgs-for-circt`.
-  circt = self.inputs.nixpkgs-for-circt.legacyPackages."${final.system}".circt;
+  circt = self.inputs.nixpkgs-for-circt.legacyPackages."${final.system}".circt.overrideAttrs (old: rec {
+    version = "1.73.0";
```
sequencer, line 18: just go to 1.74.0

nix/t1/om.nix
```diff
@@ -0,0 +1,25 @@
...
+stdenvNoCC.mkDerivation {
+  name = "t1-${mlirbc.elaborateConfig}-${mlirbc.elaborateTarget}-om";
+
+  nativeBuildInputs = [ circt omreader ];
+
+  buildCommand = ''
+    omreader ${lib.escapeShellArgs omReaderArgs}
+  '';
+
+  meta.description = "TODO.";
```
sequencer, line 24: finish this TODO please

sequencer (changes requested):

nix/t1/om.nix
SpriteOvO: Done.

nix/overlay.nix
SpriteOvO: Blocked by NixOS/nixpkgs#308512.

sequencer (approved):
wait for ci

sequencer:
This bump doesn't include the 1.74 version of circt...

sequencer (approved):

sequencer:
Cc @SharzyL this api is used to handle all metadatas our c header and emu should use it.

PR #496, opened 2024-04-02
Use D/I for T1

sequencer:


ipemu/src/TestBench.scala
```diff
@@ -5,19 +5,70 @@ package org.chipsalliance.t1.ipemu
...
+  //       in this PR, we should manually expose those verification-only APIs via probe, and connect to T1.
+  //       It can keep API stable and documented.
+  // It should read signals from probe as a new signal here.
+  dut.probeStage0EnqueueReady.zipWithIndex.flatMap(lane =>
+    lane._1.zipWithIndex.map { slot =>
+      val data = Wire(chiselTypeOf(slot._1)).suggestName(s"Lane${lane._2}Slot${slot._2}ProbeStage0EnqueueReady")
+      data := probe.read(slot._1)
+      data
+    }
+  )
```
sequencer, line 73: Use API like this to read signals from Probe IO.
For signals sending to `tlPeek`, `peekIssue`, and other DPI signals.
read it from `dut` and send to via `:=`

PR #499, opened 2024-04-02
Add Physical Design Report for CI.

Avimitin:


.github/workflows/test.yml
```diff
@@ -0,0 +1,33 @@
+name: Test Report
```
sequencer, line 1:
```suggestion
name: Physical Design Report
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,33 @@
...
+on: [pull_request]
```
sequencer, line 2: trigger by tag, maybe: Require Physical Design

PR #505, opened 2024-04-07
[ci] Add daily regression test workflow

Avimitin:
The new regression test workflow CI will try to run all performance cases and other cases that doesn't belongs to current test plan at 6:30 AM UTC+8. It should behaves like the old post PR workflow.


sequencer:
We need to consider add support to https://github.com/camel-cdr/rvv-bench.

PR #514, opened 2024-04-09
Draft Instruction Object Model

sequencer:


t1/src/decoder/InstructionDocumentation.scala
```diff
@@ -0,0 +1,425 @@
...
+// SPDX-FileCopyrightText: 2022 Jiuyang Liu <liu@jiuyang.me>
+
+package org.chipsalliance.t1.rtl
+
+import org.chipsalliance.rvdecoderdb.Instruction
+
+/** Generate documentation for each instructions for T1.
+  * The documentation should contain the behavior for instruction in a specific configuration in T1.
+  */
+case class InstructionDocumentation(instruction: Instruction, t1Parameter: T1Parameter) {
```
sequencer, line 11: This is how to start to review the decoder: adding documentation to each instructions. Describing the lifetime of each instruction in a specific type of T1.
After finish this review, a clear documentation will be a artifact.

t1/src/decoder/T1DecodePattern.scala
```diff
@@ -0,0 +1,114 @@
...
+  override def bitPat: BitPat = BitPat("b" + instruction.encoding.toString)
+
+  def documenation: String = InstructionDocumentation(instruction, t1Parameter).toString
+
+  // Below is the Scala in-memory attributes queried from DecodeTable.
+  def isVector = instruction.instructionSet.name == "rv_v"
+  // TODO: we can even use a Enum here.
+  def useLaneExec: String = ???
+
+
```
sequencer, line 95: Add attribute like this.

t1/src/decoder/InstructionDocumentation.scala
sequencer: we may also add attributes here parameters? I'm not sure.

sequencer:
@Lucas-Wye This will be 2 weeks work, let's finish it during next two weeks, It should be the first round of GSoC project.

Issue #521, opened 2024-04-14
IP Emulator should set CSR for performance evaulation.


sequencer:
IP Emulator is designed for simulate vector instructions w/o a scalar core, thus we use spike as scalar core. For now, each vector commits only used for cosim, and won't change the state of scalar core.

If we want to do some performance benchmark with ip emulator, we should set `mcycle` with simulation cycles.

This is not grantee to be accurate for program performance, but should be fine for an intensive vectorized workload.

In the future, we should also think about adding `Zihpm` support for T1, here is the roadmap:
- Start to perf with chisel `Probe` API, it will be lowered to Verilog XMR, without adding hardware overhead to RTL.
- Sort out all important events we need to care about, schedule hpm counter for these events.
- Change these `Probe` to a real hardware IO, and maintain performance counter on the `Sequencer`.
- Expose these CSR(maybe hardwired) to scalar CSR Module(we also need to refactor RocketCSR).

PR #522, opened 2024-04-14
[cosim] support cycle counters (#521)

SharzyL:


ipemu/csrc/vbridge_impl.cc
```diff
@@ -377,34 +380,44 @@ std::optional<SpikeEvent> VBridgeImpl::spike_step() {
...
       break;
     case PC_SERIALIZE_AFTER:
       break;
     default:
-      FATAL(fmt::format("SpikeStep: invalid pc: {:08X}", pc));
+      FATAL(fmt::format("SpikeStep: invalid new_pc: {:08X}", new_pc));
     }
   }
 
+  // spike does not bump mcycle by itself, so do it manually
```
Clo91eaf, line 418: Maybe we should bump `spike_cycles` first and then record it to `mcycle`, or initialize `spike_cycles` with 1.

SharzyL: You mean `mcycle` should start with 1? I did not expect that. May I ask for reasons?

Clo91eaf: sorry, I misunderstand the difference between `spike_cycle` and `mcycle`, you are right.
review passed.

Clo91eaf (approved):
🥳passed!

ipemu/csrc/vbridge_impl.cc
```diff
@@ -361,12 +361,15 @@ void VBridgeImpl::getCoverage() { return ctx->coveragep()->write(); }
 
 std::optional<SpikeEvent> VBridgeImpl::spike_step() {
   auto state = proc.get_state();
-  reg_t pc = state->pc;
-  auto fetch = proc.get_mmu()->load_insn(pc);
-  //  VLOG(1) << fmt::format("pre-exec (pc={:08X})", state->pc);
+
+  state->mcycle->write((int64_t) get_t() + spike_cycles);
```
sequencer, line 365: maybe:
```
  state->mcycle->write((int64_t) get_t() / 10 + spike_cycles);
```
get_t() is the tick, which is should be `cycle / 10`

PR #525, opened 2024-04-17
[script] use graalvm for t1-helper

Avimitin:


script/build.sc
```diff
@@ -17,4 +17,22 @@ object script extends RootModule with ScalaModule {
   def scalaVersion = v.scala3
   def scalacOptions = Seq("-new-syntax")
   override def ivyDeps = Agg(v.mainargs, v.oslib, v.upickle)
+
+  // This target assume that java executable is from GraalVM and the native-image executable is exists.
+  // Also require a working GCC
```
sequencer, line 22: I prefer not adding this in mill. But package it in nix.

sequencer (commented):
I doubt if Panama would work or not. Let’s see

script/build.sc
```diff
@@ -9,12 +9,29 @@ import mill.scalalib.scalafmt._
...
+    val outputArgs = Seq(
+      "-jar",
+      jar.path.toString(),
+      (T.dest / "out.elf").toString()
+    )
+    os.proc(
+      Seq("native-image") ++ nativeImageArgs ++ outputArgs
+    ).call(stdout = os.Inherit, stderr = os.Inherit)
+    PathRef(T.dest / "out.elf")
+  }
```
sequencer, line 36: Move these to Nix, mill should only release assembly `jar`, while using graalvm to build them to elf.

This shared scripts can be used for migrating our other jars to native for speed up the elaboration.

PR #542, opened 2024-04-23
[difftest] draft offline difftest

Clo91eaf:
- [x] add libspike-interfaces nix support
- [x] add nix-rust support for offline difftest
- [x] add more check with log file
- [ ] replace online difftest with offline difftest 
- [ ] use offline difftest in ci


difftest/offline_difftest/src/difftest.rs
```diff
@@ -0,0 +1,126 @@
...
+	shifterModuleParameters: Vec<(LogicModuleParameter, Vec<u32>)>,
+	mulModuleParameters: Vec<(LogicModuleParameter, Vec<u32>)>,
+	divModuleParameters: Vec<(LogicModuleParameter, Vec<u32>)>,
+	divfpModuleParameters: Vec<(LogicModuleParameter, Vec<u32>)>,
+	floatModuleParameters: Vec<(LogicModuleParameter, Vec<u32>)>,
+	otherModuleParameters: Vec<(OtherModuleParameters, Vec<u32>)>,
+}
+
+#[derive(Deserialize, Debug)]
+pub struct Parameter {
```
sequencer, line 66: In the future it won’t be read from the json config. But the refactor won’t be large.

sequencer: Those metadatas will use omreader reading from the mlirbc

PR #555, opened 2024-05-03
add llama perf test case

SharzyL:
- **[nix] use single-float abi compilerrt**
- **[tests] fix t1.ld DDR map**
- **[emurt] support uart/print, export header**
- **[cases] add perf.llama**


tests/perf/llama/default.nix
```diff
@@ -0,0 +1,48 @@
...
+let
+  checkpoint_bin = fetchurl {
+    url = "https://huggingface.co/karpathy/tinyllamas/resolve/main/stories15M.bin";
+    sha256 = "sha256-zVkGRNljhnorbloRB/UfrWY8QdecFJ++y7sflfqB9Jo=";
+  };
+
+  tokenizer_bin = fetchurl {
+    url = "https://github.com/karpathy/llama2.c/raw/b3c4b6c3c4bbff42e5211293280307019368ccb5/tokenizer.bin";
+    sha256 = "sha256-UKUu+CLunoPeXOnQvgoCWnc9AZQ39Ytf+dyvsGPs42E=";
+  };
```
sequencer, line 16: oh my

tests/perf/llama/run.c
```diff
@@ -0,0 +1,941 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <ctype.h>
+#include <time.h>
+#include <math.h>
+#include <string.h>
+#include <fcntl.h>
+#if defined _WIN32
+    #include "win.h"
+#else
```
sequencer, line 12: remove it?

tests/perf/llama/run.c
```diff
@@ -0,0 +1,941 @@
...
+    Config config; // the hyperparameters of the architecture (the blueprint)
+    TransformerWeights weights; // the weights of the model
+    RunState state; // buffers for the "wave" of activations in the forward pass
+    // some more state needed to properly clean up the memory mapping (sigh)
+    int fd; // file descriptor for memory mapping
+    float* data; // memory mapped data pointer
+    ssize_t file_size; // size of the checkpoint file in bytes
+} Transformer;
+
+void malloc_run_state(RunState* s, Config* p) {
```
sequencer, line 84: we may need to allocate them to specific memory(SRAM) range in the future.

sequencer:
nice!

tests/perf/llama/run.c
SharzyL: Just following the upstream but totally ok to remove

https://github.com/karpathy/llama2.c/blob/master/run.c#L10-L12

sequencer:
please add documentation to run the test case.

sequencer:
RTL still has some bug running llama. Will be fixed in the following PRs

sequencer:
The llama still hang at spike side, in order to not block others commit, Let get this into master branch, and debug it later.

PR #566, opened 2024-05-09
[rtl] Handle chaining check with different relative positions of vd and vsx.

qinjun-li:


sequencer:
Need to find a minimal test case to minimal illustrate this bug

PR #569, opened 2024-05-10
[cases] refactor case builders

SharzyL:
- **[cosim] less verbose call-frame-not-found logging**
- **[cases] remove unused uart test files**
- **[cases] refactor case builders**


tests/builder.nix
```diff
@@ -24,17 +27,22 @@ stdenv.mkDerivation (self: rec {
...
-    "-g"
-    "-O3"
-  ];
+  NIX_CFLAGS_COMPILE =
+    let
+      march = (if isFp then "rv32gc_zve32f" else "rv32gc_zve32x")
+        + (if vLen >= 32 then "_zvl${toString (lib.min 1024 vLen)}b" else "");
+    in
+    [
+      "-mabi=ilp32f"
```
sequencer, line 36: if no float, how about `ilp32`? and should we be the `ilp32d`?

tests/builder.nix
```diff
@@ -24,17 +27,22 @@ stdenv.mkDerivation (self: rec {
...
-    "-mcmodel=medany"
-    "-fvisibility=hidden"
-    "-fno-PIC"
-    "-g"
-    "-O3"
-  ];
+  NIX_CFLAGS_COMPILE =
+    let
+      march = (if isFp then "rv32gc_zve32f" else "rv32gc_zve32x")
+        + (if vLen >= 32 then "_zvl${toString (lib.min 1024 vLen)}b" else "");
```
sequencer, line 33: i think vLen is always >= 32? and we can make an asseration here.

PR #571, opened 2024-05-10
[doc] update nix usage for building docker image

SyedHassanUlHaq:
I added the following section to the script 

- Installing Docker.
- How to build the docker image locally.
- How to use the Docker image.
- How to bind the current path to /workspace.


nix/t1/release/doc.typ
```diff
@@ -37,20 +37,80 @@
 
 Scalar core cannot access Vector DDR/SRAM, for, users need to access corresponding memory banks via vector load store instructions.
 
-== How to use the Docker image
+== Installing Docker
```
sequencer, line 40:
```suggestion
== Installing Docker for Debian-based systems.
```

sequencer (commented):
cc @Avimitin and @Lucas-Wye for review.

sequencer:
Thanks for your contribution, seems you have already built the Docker container via nix. And the code owner @Avimitin for review.

Is there any official installing guide for docker? I personally prefer ref the install flow to there. WDTY?

SyedHassanUlHaq:
Yes, reference to the official installing guide for docker would be better to not make it Platform Bias. I've added it

Avimitin (commented):
No need for sudo to just copy the file to /tmp, and why we need to hash and copy file? Simple nix build and docker load should do all the jobs.

Avimitin:
Also, IMO I don't want nix workflow included in release document. The release document is for downstream docker user, and nix usage is for T1 developer.

nix/t1/release/doc.typ
```diff
@@ -37,20 +37,49 @@
...
 
 #show raw.where(lang: "t1-docker"): it => {
   raw(lang: "bash", it.text.replace("${config}", config.name))
-}
+}  
+```t1-docker
+closure="$(nix build -L '.#t1.${config}.release.docker-layers.final-image' --no-link --print-out-paths)"
+echo "path: $closure"
+sudo cp "$closure/image.tar" /tmp/t1-${config}-image.tar
+echo "cache-key-${config}=$(nix hash file --base32 $closure/image.tar)"
```
Avimitin, line 57: What is this for? You can't just copy-pasting CI workflow without knowing its usage.

SyedHassanUlHaq: I copy-pasted these to ensure consistency with the CI workflow. But you're right, that's irrelevant if the doc is meant to be only for downstream docker users. should I keep it only for the downstream docker users or make a separate section in the document for T1 developers too, WDYT?

And also what about the doc itself being built by nix, anyone accessing the doc would still need nix.

Avimitin: No, this is irrelevant because you are echoing nonsense contents.

Avimitin: This document is going to be relased as single individual PDF, which should be accessible at repository release page. Nix usage should be documented in readme.

Avimitin:
My suggestion:

- Move the whole nix document part to readme with the following changes:
    * Remove the "hash" part, the developer building the docker image doesn't need to echo this content.
    * Remove the "copy to /tmp" part, firstly it introduces wrong sudo usage, secondly it does nothing but just increase nonsense disk IO.
- Remove the modification of the docker bind mount document, users will make their own choice if they don't need it. And I want to make sure that document is concise and informative.

docs
```diff
@@ -0,0 +1 @@
+/nix/store/anpilq4vqsw5abr08g65f9hknkqs6f0h-blastoise-typst-release-doc
```
sequencer, line 1: Please review the `git diff`

nix/t1/release/doc.typ
```diff
@@ -68,4 +68,4 @@ ip-emulator --case cases/intrinsic-matmul/bin/intrinsic.matmul.elf
 
 # Get waveform trace file
 ip-emulator-trace --case cases/intrinsic-linear_normalization/bin/intrinsic.linear_normalization.elf
-```
+```
```
sequencer, line 71: nit: none related change.

Avimitin:
Please rebase master branch and squash all commit into one.

sequencer:
Please squash to one commit.

PR #574, opened 2024-05-11
Only need to detect raw when reading.

qinjun-li:


sequencer:
Please reword the commit message.

PR #583, opened 2024-05-14
fix bench

qinjun-li:


nix/overlay.nix
```diff
@@ -22,6 +22,7 @@ in
   fetchMillDeps = final.callPackage ./pkgs/mill-builder.nix { };
   circt-full = final.callPackage ./pkgs/circt-full.nix { };
   rvv-codegen = final.callPackage ./pkgs/rvv-codegen.nix { };
+  add-determinism = final.callPackage ./pkgs/add-determinism { };  # faster strip-undetereminism
```
sequencer, line 25: Nice, T1 will use it, and wanna see it in the nixpkgs in the future.

.github/cases/blastoise/default.json
```diff
@@ -498,5 +498,16 @@
   "codegen.vfredosum_vs": 142145,
   "codegen.vfredusum_vs": 99859,
   "codegen.vfredmax_vs": 99859,
-  "codegen.vfredmin_vs": 99859
+  "codegen.vfredmin_vs": 99859,
+  "rvv_bench.ascii_to_utf16": 1541742,
+  "rvv_bench.ascii_to_utf32": 673956,
+  "rvv_bench.byteswap": 3456435,
+  "rvv_bench.chacha20": 0,
```
sequencer, line 505: strange, it seems chacha20 and poly1305 doesn't goes to the auto-vectorization path. We may need to bump clang or switch to gcc?

PR #584, opened 2024-05-15
[rtl] set decoder input to be 32 bits.

Lucas-Wye:


sequencer:
should be removed

sequencer (approved):
Good start. The OM part doesn’t go to elaboration yet. Will be used for doc and dv in the future.

PR #602, opened 2024-05-22
Slots can always be shifted when slot idle

qinjun-li:


qinjun-li:
Copy of #400, because the lane changes too much

sequencer (approved):

PR #612, opened 2024-05-27
[nix] replace streamLayeredImage to buildImage

Avimitin:
Proot have too many bugs to be fixed, it is better to use qemu to build the docker image.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,15 +27,16 @@ jobs:
       - name: Build docker image file
         id: build
         run: |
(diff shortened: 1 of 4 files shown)


sequencer:
Since we are using qemu now. How about providing multiple arch(aarch64,x86,riscv) in our release?

.github/workflows/release.yml
```diff
@@ -2,6 +2,7 @@ name: Build Docker Image
 
 # Release on any git tag
 on:
+  pull_request:
```
sequencer, line 5: how about on tagging?

Avimitin: It is on tagging, `pull_request` is for testing only.

sequencer (approved):

PR #619, opened 2024-06-03
[nix] fix up omreader missing in nix

sequencer:


SpriteOvO (approved):
I removed them because it could avoid rebuilding the whole T1 if just omreader changed. However if we revert this, we still need a way to avoid rebuilding.. It's really slow.

sequencer:
Our inhouse flow just require a monorepo of T1, I don't want to add too much code in the inhouse flow, the reason of inhouse exist is because there are some NDA under Jiuyang and different companies, as well as EDA tools, we want them to be as minimal as possible.

PR #644, opened 2024-06-19
[ci] stop using path resolve for all testcase

Avimitin:


Avimitin:
Works great, the post CI action require 37 secs instead of 58 mins now.

PR #688, opened 2024-07-26
set the resetvector(width: 64) using information from an elf

PorterLu:
Assuming the length of the resetVector does not exceed 64, it can be represented as a long long.


Clo91eaf (approved):
approved!!

PR #693, opened 2024-07-28
[difftest] add chisel fatal check when simulation quit

Clo91eaf:


difftest/online_drive/src/dpi.rs
```diff
@@ -289,12 +289,16 @@ pub(crate) fn verilator_main() {
   let argc = c_args.len() as c_int;
   let argv = c_args_ptr.as_ptr() as *mut *mut c_char;
 
-  unsafe {
-    verilator_main_c(argc, argv);
+  let verilator_ret = unsafe { verilator_main_c(argc, argv) };
+
+  if verilator_ret == 0 {
+    std::fs::write("perf.txt", format!("total_cycles: {}", get_t()))
+      .expect("fail to write into perf.txt");
```
sequencer, line 296: This function should be handled by a DPI function call at system verilog `final` block

sequencer (approved):

PR #695, opened 2024-07-29
[nix] bump everything

sequencer:


sequencer:
slow down due to compiler upgrade

PR #702, opened 2024-08-04
Bump T1 dependencies

Avimitin:
Updated: nix/t1/t1.nix


Avimitin:
Seems not required, hash doesn't change after bump.

PR #703, opened 2024-08-06
refactor ipemu probes

Clo91eaf:
- [x] refactor probe signatures to improve readability
- [x] refactor to an single Probe to avoid using Record on the [[T1Interface]].

### DIFF
--- a/t1/src/T1.scala
+++ b/t1/src/T1.scala
@@ -304,6 +304,9 @@ class T1Probe(parameter: T1Parameter) extends Bundle {
   val instructionValid: UInt = UInt((parameter.chainingSize * 2).W)
   // instruction index for check rd
   val responseCounter: UInt = UInt(parameter.instructionIndexBits.W)
+  // probes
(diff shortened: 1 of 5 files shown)


t1/src/T1.scala
```diff
@@ -1737,6 +1721,8 @@ class T1(val parameter: T1Parameter)
     !slots.last.state.sMaskUnitExecution && !slots.last.state.idle,
     indexToOH(slots.last.record.instructionIndex, parameter.chainingSize * 2)).asUInt
   probeWire.responseCounter := responseCounter
+  probeWire.laneProbes.zip(laneVec).foreach { case (p, l) => p := probe.read(l.laneProbe) }
```
sequencer, line 1724: I personally don't like read here. but anyway we can refactor it in the future.

sequencer (approved):

sequencer:
We should adopt layer in a near future.

PR #709, opened 2024-08-08
[pytorch] add matmul sample

Avimitin:
Here is an sample for running `torch.matmul`. But don't merge it before it can run vector.


Avimitin:
Closed in favor of new framework in #742 .

PR #710, opened 2024-08-08
[rocketv] Elaborate RocketChip with MSU privilege

PorterLu:
Modification:
1. only write fiom in CSR
2. add msu config
3. fix RocketTile interface


rocketv/configs/Greninja.json
```diff
@@ -3,7 +3,7 @@
     "useAsyncReset": false,
     "clockGate": true,
     "instructionSets": ["rv32_i"],
-    "priv": "m",
```
sequencer, line 6: dont change it, please add another config.

sequencer: and split the commit message.

rocketemu/src/TestBench.scala
```diff
@@ -67,6 +67,7 @@ class TestBench(generator: SerializableModuleGenerator[RocketTile, RocketTilePar
   dut.io.mtip := 0.U
   dut.io.meip := 0.U
   dut.io.msip := 0.U
+  dut.io.seip.get := 0.U
```
sequencer, line 70:
```suggestion
  dut.io.seip.foreach(_ := 0.U)
```

sequencer (approved):

PR #721, opened 2024-08-13
Add OM for VRF

sequencer:
Our new PnR infrastructure(inhouse) depends on this.


sequencer:
link to chipsalliance/chisel#4298

PR #725, opened 2024-08-16
[rocket&t1] connect vxsat.

qinjun-li:


sequencer:
We need test cases for vxsat.

PR #738, opened 2024-08-23
Move svdpi bindings to separate repo

FanShupei:
I move all SystemVerilog DPI bindings code to a separate crate, making it reusable for other projects.

The code is published to crates.io as [svdpi](https://crates.io/crates/svdpi). The code currently lives at https://github.com/FanShupei/svdpi-rs


sequencer (approved):
After chisel-nix giving a more concrete example, let's relocate the svdpi-rs to chipsalliance.

PR #739, opened 2024-08-23
[nix] add 'dev-mode' attribute to handle script

Avimitin:
It is incovenient for users to always input strings like ".#t1.blastoise.ip.cases.mlir.hello.emu-result". So here is an env to help improve the current situation. With the "dev-mode" shell, users will have many bash function to help them set the current development "mode".

For example, if users want to test "intrinsic.linear_optimisation" with VCS emulator, and compare the result with verilator-emulator,
they will have to write "nix build
.#t1.blastoise.ip.vcs-emu.cases.intrinsic.linear_optimisation", then ".#t1.blastoise.ip.verilator-emu.cases.intrinsic.linear_optimisation"

With the new script, the test flow is separate into different phrases:

* Firstly, get into the dev-mode shell: `nix develop .#dev-mode`
* Then, set the config: `set_config blastoise`
* Users can inspect RTL with simple command too: `build vcs-emu-rtl`
* To switch between verilator and VCS, they can use: `set_emulator vcs-emu` or `set_emulator verilator-emu`
* Finally, run the case: `test intrinsic.linear_optimisation`

### DIFF
--- a/nix/overlay.nix
+++ b/nix/overlay.nix
@@ -46,6 +46,7 @@ rec {
 
   t1-script = final.callPackage ../script { };
   inherit (t1-script) t1-helper ci-helper;
+  dev-mode = final.callPackage ./pkgs/dev-mode { };
 
   # stdenv for compiling rvv programs, with ilp32f newlib and clang
   rv32-stdenv = rv32_pkgs.stdenv.override {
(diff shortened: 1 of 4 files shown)


Avimitin:
Since there are bunch of changes of nix build system in rocket-t1 branch, I would love to hold the PR until the PR #717 is merged, which avoids the need to have 2 PRs for the same functionality.

Avimitin:
I would love to add this feature into #717 directly.

PR #782, opened 2024-09-13
Add evaluation cases

SharzyL:


tests/eval/ooo_wb/ooo_wb.S
```diff
@@ -0,0 +1,37 @@
...
+    la t2, test_src_end
+    sub t2, t2, t1  # t2 = remaining length
+    xor t3, t3, t3  # t3 = sum
+loop_start:
+    vsetvli t0, t2, e8, m8, ta, ma   # t0 = vl
+    vle8.v v8, (t1)
+    vadd.vi v8, v8, 1
+
+    # scaler workload
+    lw t4, 0(t1)
```
sequencer, line 16: `lw t4, 0(t1)` will block `vle8.v v8, (t1)`

sequencer: V-V(mem)
indexed load to `v1`, then execute some how vector instructions not depending on `v1`, then use `v1` for the final exec, and start another round of indexed load

V-V(exec)
younger instr can retire before older instr, i haven't come up with a good example...

V-S
can schedule non-dependent scalar code in a large convoy of vector instructions, e.g. vld, add, vadd, sub, vmul, mul ...

PR #801, opened 2024-10-14
[chore] move workflow readme under .github/workflow

Avimitin:
I don't know why but GitHub think .github is more important than project root.

https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-readmes#about-readmes


dramforever (approved):
LGTM

PR #815, opened 2024-10-22
[RFC] DisplayDevice for t1rocket emulation

OceanS2000:
Depends on #800 , this PR contains followup DisplayDevice implementation.

Currently it simply dumps a fixed-size RGB buffer to disk as png upon writing to a control register. The specific interfaces are open to futher discussion.

### DIFF
--- a/difftest/dpi_t1rocketemu/src/bus/disp.rs
+++ b/difftest/dpi_t1rocketemu/src/bus/disp.rs
@@ -1,0 +1,138 @@
+use super::ShadowDevice;
+use std::fs::File;
+use std::io::BufWriter;
+use std::path::PathBuf;
+
+/// Device memory layout:
+/// `0x000_0000 - 0x1FF_0000` addressable frame buffer memory
+///   Picture is in packed RGB24 layout, row-major. See tests/disp/simple for example.
+///   OUTPUT_WIDTH * OUTPUT_HEIGHT writable, out-of-range writes will be ignored
+/// `0x1FF_0000 - 0x200_0000` control registers
+/// `0x1FF_0000`: read as frame counter, write to flush frame buffer as png
+/// `0x1FF_0004`: output dimensions
+///     0x04-06 16bits: OUTPUT_WIDTH (currently fixed at 960)
+///     0x06-08 16bits: OUTPUT_HEIGHT (currently fixed at 720)
+/// TODO: configurable dimension and color depth support?
+/// TODO: behavior emulation closer to actual LCD display?
+pub(super) struct DisplayDevice {
+  vram: Vec<u8>,
+  frame_counter: i32,
+}
+
+const DISPLAY_WIDTH: u32 = 960;
+const DISPLAY_HEIGHT: u32 = 720;
+
+const REG_START: usize = 0x1FF0000;
+const REG_FLUSH: usize = 0x1FF0000;
+const REG_DIM: usize = 0x1FF0020;
+
+impl ShadowDevice for DisplayDevice {
+  fn new() -> Box<dyn ShadowDevice>
+  where
+    Self: Sized,
+  {
+    Box::new(Self {
+      vram: vec![0u8; (DISPLAY_WIDTH * DISPLAY_HEIGHT * 3) as usize],
+      frame_counter: 0,
+    })
+  }
+
+  fn read_mem(&self, addr: usize, size: usize) -> Vec<u8> {
+    let start = addr;
+    let end = addr + size;
+
+    assert!(
+      !(start < REG_START && end > REG_START),
+      "Read burst should not cross register boundary"
+    );
+
+    if start < REG_START {
+      self.vram_read(addr, size)
+    } else {
+      self.reg_read(addr - REG_START, size)
+    }
+  }
+
+  fn write_mem(&mut self, addr: usize, data: u8) {
+    if addr < self.vram.len() {
+      self.vram[addr] = data;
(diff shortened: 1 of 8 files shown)


nix/t1/conversion/sv-to-verilator-emulator.nix
```diff
@@ -25,7 +25,7 @@ rec {
 
   verilatorFilelist = "filelist.f";
   verilatorTop = "TestBench";
-  verilatorThreads = 8;
+  verilatorThreads = 4;
```
sequencer, line 28: nit

nix/t1/mill-modules.nix
```diff
@@ -53,7 +53,6 @@ let
     passthru.editable = self.overrideAttrs (_: {
       shellHook = ''
         setupSubmodulesEditable
-        mill mill.bsp.BSP/install 0
       '';
```
sequencer, line 56: nit

sequencer (changes requested):

OceanS2000:
done

difftest/dpi_t1rocketemu/src/bus/disp.rs
```diff
@@ -0,0 +1,149 @@
...
+
+    self.vram_read(&mut result[0..vram_part_size], start, vram_part_size);
+    self.reg_read(
+      &mut result[vram_part_size..size],
+      reg_part_start - REG_FLUSH,
+      reg_part_size,
+    );
+
+    result
+  }
```
FanShupei, line 69: It's too complicated. We don't need to support one AXI transaction touching both VRAM and controll registers.

difftest/dpi_t1rocketemu/src/bus/disp.rs
```diff
@@ -0,0 +1,149 @@
...
+  fn reg_read(&self, buffer: &mut [u8], offset: usize, size: usize) {
+    let counter = self.frame_counter.to_le_bytes().into_iter();
+    let width = (DISPLAY_WIDTH as u16).to_le_bytes().into_iter();
+    let height = (DISPLAY_HEIGHT as u16).to_le_bytes().into_iter();
+    let reg_space: Vec<u8> = counter.chain(width).chain(height).collect();
+
+    let start = std::cmp::min(reg_space.len(), offset);
+    let end = std::cmp::min(reg_space.len(), offset + size);
+    buffer[0..(end - start)].copy_from_slice(&reg_space[start..end]);
+  }
```
FanShupei, line 128: We don't need to support one transaction read multiple MMIO control registers. Instead, we could requires all transactions are non-burst and naturally aligned, otherwise we report an error. A simple 'match' should suffice.

tests/disp/simple/simple.c
```diff
@@ -0,0 +1,43 @@
...
+            if (c == 1)
+                buffer[i * 960 + j].b = 0xf0;
+            else if (c > 1)
+                buffer[i * 960 + j].g = 0xf0;
+            else
+                buffer[i * 960 + j].r = 0x00;
+        }
+    }
+    *counter = 0;
+}
```
FanShupei, line 43: Add newline before EOF.

difftest/dpi_t1rocketemu/src/bus/disp.rs
```diff
@@ -0,0 +1,149 @@
...
+/// `0x1FF_0000`: read as frame counter, write to flush frame buffer as png
+/// `0x1FF_0004`: output dimensions
+///     0x04-06 16bits: OUTPUT_WIDTH (currently fixed at 960)
+///     0x06-08 16bits: OUTPUT_HEIGHT (currently fixed at 720)
+/// TODO: configurable dimension and color depth support?
+/// TODO: behavior emulation closer to actual LCD display?
+pub(super) struct DisplayDevice {
+  vram: Vec<u8>,
+  frame_counter: i32,
+}
```
FanShupei, line 18: It's better to document the pixel format and layout somewhere. IIRC, we use RGB24 format. The size of framebuffer is `ROW * COL * 3`. Each pixel is represented by 3 consecutive uint8, denoting R,G,B component respectively.

difftest/dpi_t1rocketemu/src/bus/disp.rs
OceanS2000: While I can assume all read access to be aligned and never cross register boundry, I think the software can read part of one register (for example width and height is packed in one 32-bit register) and dealing with different sizes would not lead to a cleaner implementation than current generic one.

FanShupei:
When I test it locally, unfortunately it won't pass the offline check.

Reproduction steps:
1. run `nix build -L --impure .#t1.rookidee.t1rocketemu.run.disp.simple.vcs-emu`
2. check `result/offline-check-journal`, it shows following error

```
thread 'main' panicked at offline_t1rocketemu/src/json_events.rs:208:7:
rtl data(0x1) should be equal to board data(0x0)
stack backtraces:
...
```

FanShupei:
There's no error in your code. It's a known limitation in our difftest framework.

Our offline verifier currently has no knowledge of mmio devices, it treats every read/write as regular memory access. We could add side effects for mmio read/write in our online simulation framework, but we must preserve load value axiom (the value loaded must equal to the last store to the same address). The frame counter read-back feature violates this, so the offline verifier complains loudly.

OceanS2000:
As a temporary measure I removed all read operation of `counter` from `simple.c`

FanShupei (approved):
LGTM now.

PR #788, opened 2024-09-20
[rtl] refactor mask unit.

qinjun-li:

PR #816, opened 2024-10-22
[ci] make sure all config specify in designs dir elaboratable

Avimitin:


Avimitin:
This PR is based on #788, don't merge before it.

PR #817, opened 2024-10-26
[nix] clean up impure env

Avimitin:
Yesterday I reviewed my nix code and found that this ugly part can be immediately improved. These env should not have top level binding and evaluate with empty value everytime if "--impure" does not supply. There should be value bind to them when they get evaluated, or they should never exist in our nixpkgs namespace.


Avimitin:
wat? why the performance got improved in this PR.

Avimitin:
Seems the last CI report workflow was skip due to external repository issue. So the performance improvement doesn't get recorded. https://github.com/chipsalliance/t1/actions/runs/11477882170/job/31941451053 not a big deal, ok to merge this pr.

sequencer (approved):

PR #823, opened 2024-10-31
[rtl] complete stdlib Queue and add "almost" ports

unlsycn:


stdlib/src/Queue.scala
```diff
@@ -23,79 +25,108 @@ object Queue {
...
+      "Entries must be between 1 and 1024"
+    )
 
-    val clock = Module.clock
-    val reset = Module.reset
+    val io = Wire(new QueueIO(gen, entries))
 
-    val io = Wire(new QueueIO(gen))
+    if (entries == 1) {
+      val data  = Reg(gen)
```
sequencer, line 55:
```suggestion
      val data  = if(resetMem)RegInit(0.U.asTypeOf(gen))else(Reg(gen))
```

sequencer (approved):

PR #826, opened 2024-11-03
Remove embedded ClockGen in TestBench

FanShupei:
**Proposed Design**

This PR proposes we should try move simulation control logic (clock gen, dump wave, watch dog, etc) to `TestDriver.sv`. These features are inherently coupled with simulation platform. For VCS flow, it's best to address these in verilog code.

Current Design:
```verilog
module TestBench; // top module, defined in Chisel
  ClockGen clockGen(.clock, .reset); // inlined in Chisel
endmodule
```

~Proposed Design:~
```
module TestDrvier; // top module, defined in sv
  // clock gen logic
  TestBench tb(.clock, .reset);
endmodule

module TestBench(input clock, input reset); // defined in Chisel
endmodule
```

**Benefits**

- rapid iteration cycle: currently touch everything in ClockGen requires re-elaboration, which is painfully slow.
- ~extensible to Arcilator in the future: I suppose we need a dedicated test driver for arc anyway, it's very unlike to reuse current ClockGen.~

**Future work**: I plan do refacoring of simulation control code, the current code is too messy. This is the first step.

**UPDATE**: After discusssion, we reach consensus that no need to  pull inner ClockGen to outer TestDriver. This PR only change inlined ClockGen to ExtModule.

### DIFF
--- a/nix/t1/conversion/sv-to-vcs-simulator.nix
+++ b/nix/t1/conversion/sv-to-vcs-simulator.nix
@@ -6,10 +6,13 @@
 
 { mainProgram
 , rtl
+, vsrc
 , enableTrace ? false
 , vcsLinkLibs ? [ ]
+, topModule ? null
 }:
 
+assert lib.assertMsg (builtins.typeOf vsrc == "list") "vsrc should be a list of file path";
 assert lib.assertMsg (builtins.typeOf vcsLinkLibs == "list") "vcsLinkLibs should be list of strings";
 assert lib.assertMsg (builtins.length vcsLinkLibs > 0) "vcsLinkLibs should contain at least one static library path to link";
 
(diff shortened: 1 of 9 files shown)


nix/t1/conversion/sv-to-vcs-simulator.nix
```diff
@@ -22,7 +24,12 @@ stdenv.mkDerivation rec {
   dontPatchELF = true;
   enableCover = true;
 
-  src = rtl;
+  srcs = [
+    rtl
+    vsrc
```
Avimitin, line 29: Since the following vcs command line args are supplied with full vsrc path. Do we need vsrc in source array?

FanShupei: I'm not familiar with nix. But after removing 'srcs', it also works. Are you suggesting we should remove 'srcs' here?

Avimitin: Just let src to be rtl should be OK.

sequencer:
Does verilator work now.

PR #832, opened 2024-11-05
Watchdog refactor

FanShupei:


difftest/dpi_t1emu/src/dpi.rs
```diff
@@ -236,16 +235,14 @@ unsafe extern "C" fn t1_cosim_final() {
   })
 }
 
-/// evaluate at every 1024 cycles, return reason = 0 to continue simulation,
-/// other value is used as error code.
+/// evaluate at every cycle
+/// return value:
+///   0   : continue
+///   255 : quit successfully
+///   otherwise : error
```
sequencer, line 242: Will it slowdown the simulation?

difftest/dpi_t1rocketemu/src/drive.rs
```diff
@@ -249,29 +245,24 @@ impl Driver {
   pub(crate) fn watchdog(&mut self) -> u8 {
     const WATCHDOG_CONTINUE: u8 = 0;
     const WATCHDOG_TIMEOUT: u8 = 1;
+    const WATCHDOG_QUIT: u8 = 255;
```
sequencer, line 248: Since we have the MMIO write infra, I personally prefer using a special MMIO write as exit function.

sequencer: Then watchdog is only for timeout.

t1emu/vsrc/ClockGen.sv
```diff
@@ -1,29 +1,85 @@
...
+    forever begin
+      clock = 1'b1;
+      #10;
+      clock = 1'b0;
+      #10;
+
+      cycle += 1;
+
+      begin
+        automatic byte unsigned st = t1_cosim_watchdog();
```
sequencer, line 56: I personally prefer not trigger it at each cycle.

difftest/dpi_t1rocketemu/src/drive.rs
FanShupei: The name watchdog is a bit confusing, maybe a better name is `t1_cosim_get_status`.

A special MMIO write can not reduce the complexity here. Note that we can not immediate call `$finish` during the mmio write DPI function, because we must wait TestBench becoming idle. If we do not poll rust side every cycle, an alternative design is expose a DPI `set_finish_flag` to rust side. Personally I don't like DPI calls jumping back and forth, so I tend to poll every cycle here.

difftest/dpi_t1emu/src/dpi.rs
FanShupei: My intuition is that a small function call every cycle won't slowdown. I could do some measurements later.

PR #865, opened 2024-11-25
Mmio Redesign

FanShupei:


difftest/dpi_t1rocketemu/src/interconnect.rs
```diff
@@ -0,0 +1,221 @@
...
+    None
+  }
+}
+
+/// Memory map:
+/// - 0x0400_0000 - 0x0600_0000 : framebuffer
+/// - 0x1000_0000 - 0x1000_1000 : simctrl
+/// - 0x2000_0000 - 0xc000_0000 : ddr
+/// - 0xc000_0000 - 0xc040_0000 : sram
+/// TODO: simctrl is inside ddr, move it elsewhere
```
sequencer, line 200: Fixed?

FanShupei: Yes, it's fixed. The inaccurate comment is removed in new commit.

PR #873, opened 2024-12-01
[ci] update pd report

Avimitin:


Avimitin:
Maybe we should add new label like `PD-T1RocketTile` to control the backend compile target?

PR #890, opened 2024-12-10
Use printf fd

FanShupei:
We may land it currently, or wait for circt support is ready, either


sequencer:
link to llvm/circt#7983

PR #926, opened 2025-01-06
[tests] add cover demo

Clo91eaf:
#837 

Add coverpoint support to the simplest mlir.hello test case as a cover demo.

- Added hello.json file at the same level as the source code to describe coverpoints.
- Modified default.nix build script to use jq to parse hello.json into URG coverage hierarchy format.

### DIFF
--- a/README.md
+++ b/README.md
@@ -334,6 +334,51 @@ $ nix build .#t1.<config-name>.<top-name>.cases.intrinsic.matmul -L
 $ ls -al ./result
 ```
 
+#### Developing Coverage
+
+To develop coverage, use the following steps:
+
+1. Write the coverpoint description file at the same level as the test case source code.
+2. Update the `default.nix` file to parse the coverpoint description file.
+
+For example, to develop coverage for the `mlir.hello` test case:
+
+tests/mlir/hello/hello.json:
+```json
+{
+  "assert": [
+    {
+      "name": "vmv_v_i",
+      "description": "single instruction vmv.v.i"
+    }
+  ],
+  "tree": [],
+  "module": []
+}
+```
+
+tests/mlir/default.nix:
+```shell
+if [ -f ${caseName}.json ]; then
+  ${jq}/bin/jq -r '[.assert[] | "+assert " + .name] + [.tree[] | "+tree " + .name] + [.module[] | "+module " + .name] | .[]' \
+      ${caseName}.json > $pname.cover
+else 
+  echo "-assert *" > $pname.cover
+fi
+```
+
+Then, you can run the test building script to check if the coverage is generated correctly:
+
+```shell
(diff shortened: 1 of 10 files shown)


sequencer:
Please add a documentation about how to add coverage points, tests.

README.md
```diff
@@ -328,6 +334,51 @@ $ nix build .#t1.<config-name>.<top-name>.cases.intrinsic.matmul -L
...
+
+For example, to develop coverage for the `mlir.hello` test case:
+
+tests/mlir/hello/hello.json:
+```json
+{
+  "assert": [
+    {
+      "name": "vmv_v_i",
+      "description": "single instruction vmv.v.i"
```
sequencer, line 352: How do you align the RTL coverpoint(from LTL) to this name?

README.md
```diff
@@ -328,6 +334,51 @@ $ nix build .#t1.<config-name>.<top-name>.cases.intrinsic.matmul -L
...
+
+1. Write the coverpoint description file at the same level as the test case source code.
+2. Update the `default.nix` file to parse the coverpoint description file.
+
+For example, to develop coverage for the `mlir.hello` test case:
+
+tests/mlir/hello/hello.json:
+```json
+{
+  "assert": [
```
sequencer, line 349:
```suggestion
  "cover": [
```
?

Clo91eaf: The `assert` here is the keyword of hierachy file. jq parses this json into the corresponding hierachy file. 
The detailed rules are mentioned on page 4 of Chapter 2 of the Coverage Technology Reference Manual of VCS. Here is a copy:
> Examples
> 
> -assert my_assert  
> 
> Disables all assertions with name my_assert in the full design.
> 
> -assert A*  
> 
> Disables all assertions whose name starts with A in the full design.

sequencer (approved):

README.md
Clo91eaf: There are actually two rules here: 
1. When Chisel lowers to system verilog, it will replace `.` in the [label of LTL](https://github.com/chipsalliance/t1/blob/cd9035e70809fa69bc63d215695f39dc87f61249/t1/src/T1.scala#L1049-L1054) in the Chisel code with `_`.
2. VCS enables the assertions based on wildcard matching of the name in the full design:
If a full hierarchical name is provided, such as A.C, vcs will determine the matching label. If only the last name is provided, such as C, vcs will still match the same label A.C. It is worth noting that if If there is another name, such as B.C, then vcs will match both A.C and B.C. And Some wildcard operation are also allowed, refer to `the Coverage Technology Reference Manual of VCS`
So the 'vmv_v_i' here actually corresponds to the full hierachy name 'TestBench.dut.verification.vmv_v_i'

PR #936, opened 2025-01-09
[tests] replace perf.json total_cycles value when instrument found

Avimitin:

### DIFF
--- a/.github/workflows/mmio-check.yml
+++ b/.github/workflows/mmio-check.yml
@@ -1,0 +1,78 @@
+name: T1 MMIO Check
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - ready_for_review
+      - labeled
+env:
+  USER: runner
+
+# Cancel the current workflow when new commit pushed
+concurrency:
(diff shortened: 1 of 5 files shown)


Avimitin:
This PR is intended to correctly calculate the actual cycle count for the rvv-bench project. However, each test in rvv-bench will start Scalar, Auto Vectorization and RVV tests, it is tedious to insert program instrument to all tests. Besides, adding a patch to benchmark tests that cannot be upstreamed is not what we expected.

So this PR can be merged after CI pass, but currently we will not use this feature for our test cases.

FanShupei:
Personally I suggest add data instead of replacing it. The file `perf.json` actually record various metadata for the simulation of DPI side. If the name of fields in perf.json is ambiguous, free free to change it at https://github.com/chipsalliance/t1/blob/65ae0fa4317ebea48f2e952cf030de3898d3186a/difftest/dpi_common/src/util.rs#L3C1-L22C2

Maybe we change orignal `total_cycle` to `sim_total_cycle`, add the new cycle as `perf_cycle`?

sequencer (approved):

PR #946, opened 2025-01-18
[WIP] [tests] add test framework new design RFC

Avimitin:
I am hungry, draft this for now.

### DIFF
--- a/tests/designs.md
+++ b/tests/designs.md
@@ -1,0 +1,97 @@
+---
+start-date: 2025-01-18
+author: Avimitin
+---
+
+# Summary
+[summary]: #summary
+
+This document is here to describe the new Nix build system designs for T1 test
+cases category. It aims to provide a tests framework that integrate all the T1
+compiler, emulator, linker script...etc to one environment, and any developer,
+even those who doesn't know how Nix works can easily get their code working on
+T1 emulator.
+
+# Motivation
+[motivation]: #motivation
+
+T1 RTLs is now stable and the needs for more contributor to support more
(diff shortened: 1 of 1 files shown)


sequencer:
- The `verialtor` environments is designed as emulator for software engineer and the commercial custom early evaluations. It's not designed for **verification**, thus it's provided along with the docker with necessary software toolchains inside it.
- The `vcs/ncsim` environments are designed as **verification** environments which is our signoff standard. It's not going to be useful in the SW world, and only run at our customers physical machines. I'm even not sure if it should be open-sourced. I currently triggered by GitHub Action, but maybe take down and inhouse in the future.
- CI and engineer UX are multiple things we may mixed them, I sort them out, and please take them into consideration.
  - Reproducibility and scaling up for parallelism are first concerns for CI, that's the best thing you have done with Nix framework.
  - RTL Engineer UX should be able to access the waveform after changing RTL, the compile and simulation should happen in the remote machine and the designer only require the artifacts of simulation for debugging.
  - DV Engineer UX is similar to RTL and should be able to simply add test cases and get the reports immediately(the vcs/ncsim emulator should be fetched from nix cache), the coverage merging flow also needs to be re-triggered to get the entire coverage report.
  - SW UX should be able to run their own program under the `verilator` emulation flow, they don't concern the correctness(consider the result are always correct), if they observe the RTL bug, they should be able to submit a standard bug report, and we should have a standard flow to report to the bug to DV Engineer and create a new case for debugging.

Notice we don't have a good performance evaluation report yet... Which should be designed by @FanShupei after his on-boarding.

PR #942, opened 2025-01-14
[release] fix docker image builder

Avimitin:

PR #947, opened 2025-01-19
Fix t1rocketemu for verilator build

FanShupei:
The root cause is clocked dpi call in chisel behave as if a DFF, but it has no way to react to reset signal.

Since we currently only use sync reset, we could actually emulate it in DPI side. reset is passed as an additional parameter to DPI and DPI could set return value to zero when reset, effectively clear the register when reset is asserted.

Note in AXI read, the DPI return register has no control signal. Thus we just leave the content undetermined, to make DPI code cleaner.


FanShupei:
Close as included in #942

PR #948, opened 2025-01-19
Bump T1 dependencies

Avimitin:
Updated: nix/t1/dependencies/_sources/generated.json
nix/t1/dependencies/_sources/generated.nix


sequencer:
The bump is failed because of chipsalliance/chisel#4616
T1 should swap to fine-grand control to ivy dependency and use nix to control each piece.
Ask @Avimitin to get the new feature in before this Friday.

PR #950, opened 2025-01-20
[nix] seperate verilating and linking for better caching

FanShupei:
To cache better, the build of verilator-emu is split into 2 steps:

1. build verilated lib.
2. link verilated lib with dpi lib, get the final executable.

The advantage is that change of dpi code will not invalidate verilated lib

### DIFF
--- a/nix/t1/conversion/link-verilator-with-dpi.nix
+++ b/nix/t1/conversion/link-verilator-with-dpi.nix
@@ -34,6 +34,6 @@ stdenv.mkDerivation (rec {
   '';
 
   passthru = {
-    enableTrace = verilatorLib.enableTrace;
+    inherit (verilatorLib) enableTrace;
   };
 })
--- a/nix/t1/conversion/link-verilator-with-dpi.nix
+++ b/nix/t1/conversion/link-verilator-with-dpi.nix
@@ -1,0 +1,39 @@
+{ lib
+, stdenv
+, zlib
+}:
+
+{ mainProgram
+, verilatorLib
+, dpiLibs
+}:
+
+assert lib.assertMsg (builtins.typeOf dpiLibs == "list") "dpiLibs should be a list of file path";
+
+stdenv.mkDerivation (rec {
+  name = mainProgram;
+  inherit mainProgram;
+
+  dontUnpack = true;
+
+  propagatedBuildInputs = [ zlib ];
+
+  ccArgs = [
+    "${verilatorLib}/lib/libV${verilatorLib.topModule}.a"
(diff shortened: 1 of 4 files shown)


nix/t1/conversion/link-verilator-with-dpi.nix
```diff
@@ -0,0 +1,39 @@
...
+
+stdenv.mkDerivation (rec {
+  name = mainProgram;
+  inherit mainProgram;
+
+  dontUnpack = true;
+
+  propagatedBuildInputs = [ zlib ];
+
+  ccArgs = [
```
Avimitin, line 21: Try `NIX_CFLAGS_COMPILE`, which will be automatically read and handle in stdenv.

nix/t1/conversion/link-verilator-with-dpi.nix
```diff
@@ -0,0 +1,39 @@
...
+  ];
+
+  buildCommand = ''
+    mkdir -p $out/bin
+
+    $CXX -o $out/bin/$mainProgram ${lib.escapeShellArgs ccArgs}
+  '';
+
+  passthru = {
+    enableTrace = verilatorLib.enableTrace;
```
Avimitin, line 37: nitpits: `inherit (verilatorLib) enableTrace;`

nix/t1/t1.nix
```diff
@@ -117,16 +117,26 @@ forEachTop (topName: generator: self: {
...
+  };
+
+  verilator-emu = t1Scope.link-verilator-with-dpi {
+    mainProgram = "${topName}-verilated-simulator";
+    verilatorLib = self.verilator-emu-lib;
+    dpiLibs = [ "${self.verilator-dpi-lib}/lib/libdpi_${topName}.a" ];
+  };
+  verilator-emu-trace = t1Scope.link-verilator-with-dpi {
     mainProgram = "${topName}-verilated-trace-simulator";
+    verilatorLib = self.verilator-emu-trace-lib;
```
Avimitin, line 138: Since the `link-verilator-with-dpi` function derives `makeOverridable`, here we can use `verilator-emu.override { mainProgram = "${topName}-verilated-trace-simulator"; };`

nix/t1/conversion/link-verilator-with-dpi.nix
FanShupei: Here all args are inputs, not flags. And we are actually performing linking here. I feel a bit wired to pass them as CFLAGS.

Avimitin (approved):

PR #962, opened 2025-01-26
NTT Evaulation

sequencer:
- **[eval] add build**
- **[eval] add eval.ntt***


sequencer:
@Avimitin Please add CI to run all evaluation cases under `benchmark_dlen*_vlen*` for performance benchmark.

sequencer (commented):
I think we should do some thing like this:
- For `ntt_*_main.c` it should detect the `VLEN` at runtime and delegate to different algorithm.
- For `static const int arr[xxx]` and its layers, a script should be added to generate those these vector(maybe from `ref.ws`?) rather than hardcode.
- Neither do I like `gen_data.py`, it should also be rewritten into `ws` for simplicity.
- How does `twiddle` get calculated? From `gen_vector_ntt_tw.py`? I think this python should be rewrite into ws as a function, and codegen corresponding headers.

Please clarify the build stage flows:
1. Put all user inputs(maybe `arr`?) into a `csv` or `json` file.
2. At the compile time, a `ws` should read the `csv` and generate `data` and `twiddle`, the output should be store as `json` and `csv`
3. A codegen python/Scala should read output from `ws` and generate a corresponding C header file.
4. The test entry should include the C header file and execute to get the result.

Some random thing I can think about:
- How do software programmer toggle performance counters though C-API?
- What's the SW engineer API for benchmark? I think we may need a generic post-evaluation flow for the benchmark.

SharzyL (approved):

PR #963, opened 2025-01-26
Simplify offline checker

FanShupei:
In this PR, I change how ISA is passed from simulation to offline checker. Now it's recorded in `perf.json` and offline directly reads from it. Now 'perf.json' is actully 'sim_metadata.json', but I just keep the name because the name is already used everywhere.

Since now offline checker read everything from perf.json, no longer necessary to use different offline for t1emu and t1rocketemu. So I just combine them.

It simplifies build process and prevent potential errors. (e.g. docker users accidentally use the wrong version to run offline)

Usage for new offline checker: 
```
offline \
  --perf-json <PERF_JSON> \
  --log-file <RTL_EVENT_JSONL> \
  --elf-file <ELF_FILE>
```

Note: DPI will record ELF_FILE in perf.json if its path is absolute ,thus ELF_FILE could be omitted for offline in this case.


FanShupei:
@Avimitin Since it changes offline comcmand line usage, make sure CI and t1-helper are not regressed before merge.

Avimitin:
I think we should take a bold step forward and drop the `perf.json` naming, then fix everything related to this name.

FanShupei:
Change: renaming across code base
- `offline` -> `t1-sim-checker`
- `perf.json` -> `sim_result.json`

Avimitin (approved):

Avimitin (approved):

PR #967, opened 2025-02-06
[nix] add custom bundler

Avimitin:
Most of our derivation definitions have no pname attribute, and official
bundlers doesn't support bundle those derivations. This commit add my
fork as input to fix this issue. Users can run
`nix bundle --bundlers . <attribute>` to bundle a derivation.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -18,6 +18,30 @@
         "type": "github"
       }
     },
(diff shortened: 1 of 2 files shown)


Avimitin:
Upstream as https://github.com/nix-community/nix-bundle/pull/119.

PR #971, opened 2025-02-14
Fix btb

donaldkuck:
[RTL]Fix BTB: update_target is from btb_update


sequencer:
Thanks for your contribution! Ask @qinjun-li for review

PR #972, opened 2025-02-15
Bump nixpkgs

Avimitin:


sequencer (approved):
LGTM, but need to debug the reason of slowing down.

PR #976, opened 2025-02-18
Async memory DPI

CircuitCoder:


FanShupei:
I feel the current Device/Crossbar model is a bit messy. Here I propose a slightly different design, based on the following design principles:

- **make simple devices indeed simple**: a simple device may always be ready and always immediately response for request. Crossbar may further add a fixed latency 
- **make DRAM model simple**: the DRAM model could focus on how to tick dramsim, forget everything about AXI ordering. AXI ordering requirements will be handled by Completer/Crossbar.

The device model mimics AXI in hardware, but does have notable differences. Hope the differences are acceptable since out current task is just bridging dramsim with our RTL.
- **no timing modeling for burst**: the transaction always happen at single cycle.
- **no back pressure in response**: device may ack requests at any time in arbitrary order. If FIFO behavior is required, it may be emulated by a reorder buffer in Completer`

This is currently a proposal, especially lacks details in Crossbar design. A major difference from your current code is that I try to move Completer out of Device.

```Rust
// ReqToken is opaque to device, intentionally !Copy
struct ReqToken(u32);

// Inner is shared between Crossbar and Device
// Inner may contain a `VecDeque` as a reorder buffer
#[derive(Clone)]
struct Completer {
    inner: Rc<RefCell<Inner>>,
}

// Interfaces for Device
impl Completer {
    pub fn ack_read(&mut self, time: u64, token: ReqToken, data: &[u8]);
    pub fn ack_write(&mut self, time: u64, token: ReqToken);
}

// Interfaces for Crossbar
impl Inner {
    fn is_head_completed(&self) -> bool;
    
    // pop from head ...
    // alloa req token ...
}


trait Device: Any + Send {
    // Each Device is a associated with a Completer,
    // given by Crossbar as a param of constructor
    // fn get_completer(&self) -> Completer
    
    // Crossbar may not tick it if no transaction in flight
    fn tick_and_check_ready(&mut self, time: u64) -> bool;
    
    // Crossbar shall ensure the device is ready, otherwise panic
    // A simple device may ack the device immediately
    fn req_read(&mut self, time: u64, token: ReqToken, addr: AddrInfo);
    fn req_write(&mut self, time: u64, token: ReqToken, addr: AddrInfo, data: &[u8]);
}

// formerly `AddressSpace`
pub struct Crossbar {
    inner: Rc<RefCell<Inner>>,
    devices: Vec<DeviceEntry>,
    // ...
}

// impl Device CrossBar {
//   CrossBar itself could be a Device, like AXI
//   ...
// }
```

Avimitin (changes requested):
Symlinks seems to be accidentally added, please remove them.

CircuitCoder:
> Symlinks seems to be accidentally added, please remove them.

Should be fixed, thanks

difftest/dpi_t1rocketemu/src/dpi.rs
```diff
@@ -374,11 +277,20 @@ unsafe extern "C" fn t1_cosim_init(
 
   let scope = SvScope::get_current().expect("failed to get scope in t1_cosim_init");
 
+  use std::io::Write;
+  let ds3_cfg = include_bytes!("dramsim3-config.ini");
```
sequencer, line 281: pass string from SV plusargs.

sequencer: cc @FanShupei

difftest/dpi_t1rocketemu/src/dpi.rs
```diff
@@ -374,11 +277,20 @@ unsafe extern "C" fn t1_cosim_init(
...
+  ds3_cfg_file.write(ds3_cfg).expect("Unable to write DRAMsim3 configuration temp file");
+  let (_, ds3_cfg_path) =
+    ds3_cfg_file.keep().expect("Unable to persist DRAMsim3 configuration temp file");
+
   let args = OnlineArgs {
     elf_file: elf_file.get().to_str().unwrap().into(),
     dlen: dlen as u32,
     vlen: vlen as u32,
     spike_isa: spike_isa.get().to_str().unwrap().into(),
+    dramsim3_cfg: Some(ds3_cfg_path),
```
sequencer, line 293: I this here should not be a optionals dramsim will always linking.

difftest/dpi_t1rocketemu/src/dramsim3-config.ini
```diff
@@ -0,0 +1,68 @@
+# DDR4_8Gb_x8_3200 from DRAMsim3 upstream
```
sequencer, line 1: How about placing this file in a better place, it should be a configuration. cc @Avimitin

difftest/dpi_t1rocketemu/src/drive.rs
```diff
@@ -37,6 +41,218 @@ pub struct OnlineArgs {
 
   /// ISA config
   pub spike_isa: String,
+
+  /// DRAMsim3 configuartion (if any)
+  pub dramsim3_cfg: Option<PathBuf>,
```
sequencer, line 46: We need to force any downstream user to use it.
```suggestion
  pub dramsim3_cfg: PathBuf,
```

difftest/dpi_t1rocketemu/src/interconnect.rs
```diff
@@ -192,14 +591,16 @@ impl AddressSpace {
   }
 }
 
+pub const SRAM_BASE: u32 = 0x2000_0000;
+pub const SRAM_SIZE: u32 = 0xa000_0000;
```
sequencer, line 595: How does the DDR region configured? We can statically config a region for DDR, and print this region with `--help` from the emulator.

t1rocketemu/src/AXI4SlaveAgent.scala
```diff
@@ -58,189 +73,156 @@ class ReadPayload(length: Int, dataWidth: Int) extends Bundle {
...
+    true.B,
+    io.reset.asTypeOf(UInt(8.W))
+  )
+
+  /// Widen a wire to make all DPI calls' types uniform
+  def widen(wire: UInt, target: Int): UInt = {
+    require(wire.isWidthKnown)
+    require(wire.getWidth <= target)
+    wire.asTypeOf(UInt(target.W))
   }
```
sequencer, line 117: It makes me nervous...
But anyway I this is OK....
please give a tag to align the axi channel id between RTL and ID.

t1rocketemu/src/TestBench.scala
```diff
@@ -127,7 +127,7 @@ class TestBench(val parameter: T1RocketTileParameter)
   }
   instFetchAgent.io.clock := clock
   instFetchAgent.io.reset     := reset
-  instFetchAgent.io.channelId := 0.U
+  instFetchAgent.io.channelId := 2.U
```
sequencer, line 130: why this is configure with 2?

sequencer (commented):
I didn't take a deep dive into the emulator codes, leave it to @FanShupei.
need to take care of my some nitpicks:

- It made me nervous of unify the DPI calls from different channels with one DPI call, please make sure the channel ID can be simply extendable(we may add more indexed/stride AXI ports in the future)

Some thing we need to take into another consideration(maybe in the future PR):
- address region should be explicit configured, with a configuration file or with statically hardcode are both OK, SW engineers need `--help` to get the address region.
- How does the dramsim affect our verification flow? We are using printf trace-based verification, maybe we should be able to swap the RTL with sail model in the future?
- The performance is affected, we still wanna evaluate the theoretical performance with 0-latency like what we originally did, and we can evaluate the penalty of memory latency for uarch tuning.

sequencer:
The commit msg may not be difftest, just emu is enough? cc @FanShupei

t1rocketemu/src/TestBench.scala
CircuitCoder: Because the previous setting was wrong and conflicted with one of the ports (iirc highBandwidth) from vLSU

t1rocketemu/src/AXI4SlaveAgent.scala
CircuitCoder: The tag is data_width which is given as another parameter in the DPI call.

The main problem is that since now we're using the same DPI function across all channels, it has to have a uniform type. Although the DPI side doesn't care about the width of the data as long as it's wider than 64 bit and is hence presented as a pointer, but the RTL side does.

If in the future, RTL code changes cause the channel to widen and exceed the width (`target`) given here, the assertion will fail, and the target width should be raised manually.

difftest/dpi_t1rocketemu/src/interconnect.rs
CircuitCoder: Actually these two constants should be named `RAM_{BASE,SIZE}` because they also represent the DRAM address mapping. To reuse the ELF across different memory latency models, the memories have to have the same mapping.

difftest/dpi_t1rocketemu/src/drive.rs
CircuitCoder: Actually we don't. No dramsim config means that using the trivial latency model. This is a supported settings, but currently not easily settable. Switching the configuration to using plusargs should solve this.

difftest/dpi_t1rocketemu/src/dramsim3-config.ini
CircuitCoder: The planned argument for setting DRAMsim3 settings has three values:
- +dramsim3_cfg = yes: using the embedded setting, which is this one
- +dramsim3_cfg = no: disable DRAMsim3, use trivial memory latency model
- +dramsim3_cfg = <path>: Use an external DRAMsim3 setting.

The embedded configuration here are just a quality-of-life feature, and makes CI easier to use.

difftest/dpi_t1rocketemu/src/dpi.rs
CircuitCoder: I think using the trivial model is desirable in some cases?

CircuitCoder:
@sequencer Some of the question you mentioned are answered in the inline reviews. For other questions:

> How does the dramsim affect our verification flow? We are using printf trace-based verification, maybe we should be able to swap the RTL with sail model in the future?

I don't think there is an inherent difference (w.r.t. offline simulation) between sail and spike, so I think it should also be applicable? This PR doesn't touch online difftest.

> The performance is affected, we still wanna evaluate the theoretical performance with 0-latency like what we originally did, and we can evaluate the penalty of memory latency for uarch tuning.

iirc @FanShupei did some profiling, and it shows that the increase in simulation time mostly came from the increase in cycle count itself, not from the added performance penalty from using DRAMsim3. If a faster simulation time should be recovered, it can simply use the trivial memory model after we allow configurations of memory latency from plusargs.

Also, the fake cache is currently not implemented. Adding that should also help.

FanShupei:
> The commit msg may not be difftest, just emu is enough?
@sequencer I'm OK with both.  We've used `difftest` to tag any rust code change under difftest folder for a long time. No bother to change here.

@CircuitCoder The RTL fix is merged into master, this PR needs rebasing. The following commits change nix dependenceis. If these changes are intended, please separate them into another PR.
- [difftest] Bump dependencies
- [dependencies] Bump dep

The patch set is large than usual, and reviewing may need some time. I plan to finish the review in one or two days. Stay tuned.

t1rocketemu/src/AXI4SlaveAgent.scala
```diff
@@ -58,189 +73,156 @@ class ReadPayload(length: Int, dataWidth: Int) extends Bundle {
...
+      val wqueue  = Module(new Queue(new W(parameter.axiParameter), 2))
+      val bqueue  = Module(new Queue(new B(parameter.axiParameter), 2))
+
+      awqueue.io.enq <> channel.aw
+      wqueue.io.enq <> channel.w
+      bqueue.io.deq <> channel.b
+
+      // Invoke DPI at negedge
+      // NOTICE: this block CANNOT directly write any outside reg. Only write wires (e.g. here, only writes queue IO)
+      withClock(invClock) {
```
FanShupei, line 131: `RawCLockedNonVoidFunctionCall` pass clock explicitly. No need to `withClock` here.

And please separate changes on AXI4SlaveAgent (from changes on rust code) to their own commits , with `[axi4] ...` commit message , to make the history cleaner.

FanShupei (changes requested):

difftest/dpi_t1rocketemu/src/interconnect.rs
```diff
@@ -18,26 +30,64 @@ impl AddrInfo {
...
+  pub id: u64,
+  pub payload: MemRespPayload<'a>,
+}
+
 // Caller is reponsible to ensure the following conditions hold:
 //   addr.len > 0
 //   addr.len == data.len()
 //   addr.len == mask.len() (if mask present)
 // However, since the functions are safe,
 // even if contracts violate, implementions must not break memory safety,
```
FanShupei, line 86: These comments become outdated. Remove them or move them to `AddrInfo`/`MemReqPayload`

difftest/dpi_t1rocketemu/src/interconnect.rs
```diff
@@ -62,68 +112,424 @@ pub trait RegDevice {
...
+  }
+
+  fn resp(&mut self) -> Option<MemResp<'_>> {
+    let popped = self.model.pop()?;
+    // dbg!(&popped);
+
+    // Construct MemResp
+    let payload = if popped.is_write {
+      MemRespPayload::WriteAck
+    } else {
```
FanShupei, line 517: It seems to may reorder requests. W->W & W->R order is preserved. R->R reorder is harmless.

But it does allow reordering R->W to W->R. But we are modeling AXI, such order shall be handled by managers. Seems we are exploiting this behavior.

Is my understanding correct?

FanShupei (changes requested):

difftest/dpi_t1rocketemu/src/interconnect.rs
CircuitCoder: I might not correctly understand the concern here. If what you are asking is that "where lies the responsibility for ordering concurrent overlapping R&W requests", then yes, it's the manager. More concretely, it's the iterating order of `incomplete_reads` and `incomplete_writes` in `src/drive.rs`. The device handles the requests as-is.

My understanding is that this reordering is allowed by AXI. The DPI side only need to guarantee any **consistent** order, s.t. the simulation result is deterministic. I'm even thinking of switch back to HashMap with a consistent hasher to get a little bit more coverage on potential memory request orderings.

CircuitCoder:
> > The commit msg may not be difftest, just emu is enough?
> > @sequencer I'm OK with both.  We've used `difftest` to tag any rust code change under difftest folder for a long time. No bother to change here.
> 
> @CircuitCoder The RTL fix is merged into master, this PR needs rebasing. The following commits change nix dependenceis. If these changes are intended, please separate them into another PR.
> 
>     * [difftest] Bump dependencies
> 
>     * [dependencies] Bump dep
> 
> 
> The patch set is large than usual, and reviewing may need some time. I plan to finish the review in one or two days. Stay tuned.

These two commits are splited into #981. This PR now depends on that one.

difftest/dpi_t1rocketemu/src/interconnect.rs
FanShupei: OK, I just want to confirm this. I have no concerns. I agree ensuring determinism is enough. I'm just a bit puzzled since no comments mention it. Comments like "It follows AXI ordering rules. No ordering between reads & writes are guaranteed by devices " with `MemReq` helps to clarify it.

In the future, I consider adding functions to detect R/W address overlaps from the same master, since it's very likely a bug of manager. This is not a blocker for this PR.

difftest/dpi_t1rocketemu/src/interconnect.rs
CircuitCoder: Done in 68d735c3

difftest/dpi_t1rocketemu/src/interconnect.rs
CircuitCoder: Done in 5bc77dd5

difftest/dpi_t1rocketemu/src/dpi.rs
CircuitCoder: Done in 225ade2b

FanShupei (approved):

PR #987, opened 2025-03-07
[deps] fix deps script

unlsycn:


Avimitin:
Close in favor of #990

PR #1004, opened 2025-03-25
New OMReader

unlsycn:


build.mill
```diff
@@ -14,13 +14,14 @@ import coursier.maven.MavenRepository
 
 object v {
   val scala    = "2.13.16"
```
sequencer, line 16: Do we still check the format of build file?

nix/overlay.nix
```diff
@@ -22,7 +22,25 @@ rec {
 
   # Override "nixpkgs" circt with "nixpkgs-for-circt".
   # To update the "nixpkgs-for-circt" input, run `nix flake lock --update-input nixpkgs-for-circt`.
-  circt = self.inputs.nixpkgs-for-circt.legacyPackages."${final.system}".circt;
+  circt = self.inputs.nixpkgs-for-circt.legacyPackages."${final.system}".circt.overrideAttrs (prevAttrs: {
```
sequencer, line 25: Here introduce some unsafe codes, in the future we need to consider ask T1 to depends on zaozi with its Scala rewritten firtool to align the toolchain.

omreader/src/OMReader.scala
```diff
@@ -0,0 +1,29 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2025 Yuhang Zeng <unlsycn@unlsycn.com>
```
sequencer, line 2: Upstream this file to zaozi as user API, and keep your copyright there. In t1 repo all copyright are reserved by me.

omreader/src/T1OMReader.scala
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2025 Yuhang Zeng <unlsycn@unlsycn.com>
```
sequencer, line 2: please give copyright back to me ;p

omreader/src/t1/T1.scala
```diff
@@ -1,45 +1,54 @@
...
-  def march(@arg(name = "mlirbc-file") mlirbcFile: os.Path) =
-    println(new Lib(os.read.bytes(mlirbcFile)).march)
+  val vlen = t1.obj("vlen")
 
-  @main
-  def sram(@arg(name = "mlirbc-file") mlirbcFile: os.Path) =
-    new Lib(os.read.bytes(mlirbcFile)).sram.foreach(s => println(upickle.default.write(s)))
+  val dlen = t1.obj("dlen")
+
+  val instructions = t1.obj("decoder").obj("instructions").list.toSeq
```
sequencer, line 29: I think in the open source part we don't require this metadata, but anyway let's keep it here, we will clean up them in the future.

sequencer (changes requested):

build.mill
unlsycn: No, the format check requires ScalafmtModule, so we have to format them manually.

sequencer: Ok please add CI check for format in the next PR

sequencer:
It will break circt bump CI, so we require daily bump zaozi and daily bump zaozi in T1.

PR #1009, opened 2025-03-26
[tests] fix lenet

FanShupei:


FanShupei:
Close as solved in #1032.

PR #1021, opened 2025-04-01
[nix] bump riscv-vector-tests

Avimitin:
Bump revision and fix naming. Also remove the riscv-test-env submodules
bundled from upstream cuz we already vendored one in source tree.


Avimitin:
Seems like there are RTL bugs, assign this PR to @qinjun-li

PR #1023, opened 2025-04-02
Add treefmt

FantasqueX:
https://github.com/numtide/treefmt-nix

I haven't executed `nix fmt`. Do I need to add a CI check?

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -10,9 +10,10 @@
       inputs.flake-utils.follows = "flake-utils";
     };
     zaozi.url = "github:sequencer/zaozi";
+    treefmt-nix.url = "github:numtide/treefmt-nix";
   };
 
-  outputs = { self, nixpkgs, mill-ivy-fetcher, flake-utils, zaozi }@inputs:
+  outputs = { self, nixpkgs, mill-ivy-fetcher, flake-utils, zaozi, treefmt-nix }@inputs:
     let
       overlay = import ./nix/overlay.nix { inherit self; };
     in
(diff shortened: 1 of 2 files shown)


treefmt.nix
```diff
@@ -0,0 +1,5 @@
+{ pkgs, ... }:
```
Avimitin, line 1: I would like to drop this file and use plain text in flake.nix instead.

flake.nix
```diff
@@ -10,16 +10,18 @@
...
-  outputs = { self, nixpkgs, mill-ivy-fetcher, flake-utils, zaozi }@inputs:
+  outputs = { self, nixpkgs, mill-ivy-fetcher, flake-utils, zaozi, treefmt-nix }@inputs:
     let
       overlay = import ./nix/overlay.nix { inherit self; };
     in
     flake-utils.lib.eachDefaultSystem
       (system:
         let
           pkgs = import nixpkgs { inherit system; overlays = [ mill-ivy-fetcher.overlays.default zaozi.overlays.default overlay ]; };
+          treefmtEval = treefmt-nix.lib.evalModule pkgs ./treefmt.nix;
```
Avimitin, line 24: I think we can just use:

```nix
        treefmtEval = treefmt-nix.lib.evalModule pkgs {
          projectRootFile = "flake.nix";
          settings.verbose = 1;
          programs.nixfmt.enable = true;
        };
```

since we don't want to use treefmt for all language yet.

Avimitin (changes requested):

flake.nix
FantasqueX: https://treefmt.com/latest/guides/unmatched-formatters/#enforcing-strict-matching I apply this suggestion.

Avimitin:
Please use [nix] add treefmt, and fixup two commits. It is ok to force push.

Avimitin (approved):

PR #1032, opened 2025-04-07
[test] add init_array call to fix data construction

Avimitin:


FanShupei:
LGTM. 

We should add a dedicated test to ensure the C++ constructor of global variables are invoked, but that could be done in the future. Just add this to our backlog.

PR #1054, opened 2025-04-24
Update to Rust Edition 2024

FanShupei:


FanShupei:
See https://github.com/numtide/treefmt-nix/issues/343 for why I temporally disable rustfmt in nix fmt.

Avimitin (approved):

PR #1057, opened 2025-04-26
Switch to SimLog.printf

FanShupei:


nix/t1/conversion/sv-to-verilator-emulator.nix
```diff
@@ -34,7 +34,8 @@ let
       "-y"
       "$DWBB_DIR/sim_ver"
       "-Wno-lint"
-      "+define+PRINTF_FD=t1_common_pkg::log_fd"
+      "+define+PRINTF_FD=do_not_use_printf"
```
sequencer, line 37: Just delete it?
```suggestion
```

t1emu/vsrc/VerbatimModule.sv
```diff
@@ -1,14 +1,22 @@
...
   //   +t1_rtl_event_off  (optional) set to 1 to disable rtl event recording
   //   +t1_rtl_event_path            path to rtl event jsonl file
 
-  bit    log_cond;
   int    log_fd;
   string log_path;
 
   int rtl_event_off = 0;
 
+  class FileDescriptor;
```
sequencer, line 11: It seems that the FD has been managed by circt?
see: https://github.com/llvm/circt/blob/c07d31b33912b2b33d36f1f8f6312b0a4936137a/lib/Conversion/FIRRTLToHW/LowerToHW.cpp#L888-L900

FanShupei: I use "+define+__CIRCT_LIB_LOGGING" intentionally disable circt's generated implementation, for two reasons:
1. If two files both use SimLog.pritnf, it may cause duplicate definition error. (depending on how tools handle them)
2. Use our own implementation gives finer control. Now filename acts as an identifier, instead of a real filename

sequencer: OK~

nix/t1/conversion/sv-to-verilator-emulator.nix
FanShupei: If some code accidentally use `printf` instead of `SimLog.printf`, it will cause subtle bugs. This is a check for it and I believe such a check is useful.

After `PRINTF_FD` is reverted, we may find a new way to check it. Keep it now is not harmful.

PR #1072, opened 2025-04-29
t1rocketemu: handle AXI error at interconnect

FanShupei:


FanShupei:
Notable Changes:
1. Now sim interconnect handles bus error. Currently our axi interface does not support rresp & bresp, dpi code still panics on the boundary axi_pop_B & axi_pop_R. It may be addressed in future PRs.
2. When panics, we distinguish internal logic error and potential RTL bugs. The latter is explicitly noted with "SIM ERROR:" in panic message. In the future, DPI may not panic for RTL bugs, let the TB to handle it.

CircuitCoder (approved):
LGTM

PR #1089, opened 2025-05-09
[vcs] make T1_DEV define optional

Avimitin:


FanShupei (approved):

Avimitin:
This PR is not required since the Nix file is only evaluated for T1 developers.

PR #1150, opened 2025-07-08
[pokedex-difftest] add CSR log parsing

Avimitin:


pokedex/simulator/difftest/assets/pokedex-sim-event.jsonl.example
FanShupei, line 1: We should NOT include timestamp in event log. The major problem is that it makes the event log nondeterministic.

FanShupei (approved):

PR #1158, opened 2025-07-18
pokedex/simulator memory

Avimitin:


pokedex/simulator/pokedex/src/simulator.rs
```diff
@@ -359,3 +335,314 @@ impl Drop for SimulatorHandle {
...
+        addr: u32,
+        len: u32,
+    },
+    Illegal,
+}
+
+pub trait Addressable: std::fmt::Debug {
+    fn resp_bus_read(&self, offset: u32, dest: &mut [u8]) -> Result<(), BusError>;
+    fn resp_bus_write(&mut self, offset: u32, data: &[u8]) -> Result<(), BusError>;
+}
```
FanShupei, line 510: 1. may simply use `bus_read`/`bus_write`
2. MMIO read may have side effects. It should take `&mut self`
3. `Debug` super trait seems not very useful. It seems we don't want to dump the whole memory. If it's not used, we may remove it.
4. We may add `Send` trait bound somewhere, ensure it could be wrapped inside `Mutex` in the future.

pokedex/simulator/pokedex/src/simulator.rs
```diff
@@ -359,3 +335,314 @@ impl Drop for SimulatorHandle {
...
+        let length = self.memory.len() as u32;
+        if offset >= length || offset + data.len() as u32 > length {
+            return Err(BusError::OutOfMemory {
+                id: "NaiveMemoryWrite",
+                addr: offset,
+                len: data.len() as u32,
+            });
+        }
+
+        (&mut self.memory[offset as usize..offset as usize + data.len()]).copy_from_slice(&data);
```
FanShupei, line 553: no need to `&mut` manually here. If you are not so familiar with Rust, you can use `cargo clippy`, it will give you some useful suggestions.

pokedex/simulator/pokedex/src/simulator.rs
```diff
@@ -53,18 +59,20 @@ pub struct Simulator {
 }
 
 impl Simulator {
-    fn new(memory_size: usize, max_same_instruction: u8) -> Self {
+    fn new<P: AsRef<Path> + Debug>(dts: P, max_same_instruction: u8) -> miette::Result<Self> {
```
FanShupei, line 62: No need for `Debug` trait bound.
```Rust
let dts = dts.as_ref();
println("{}", dts.display()); // use `Path::display` to show path
println("{:?}, dts); // use `Debug` to show path
```

pokedex/simulator/pokedex/src/simulator.rs
```diff
@@ -359,3 +335,314 @@ impl Drop for SimulatorHandle {
...
+
+#[derive(Debug)]
+pub enum BusError {
+    OutOfMemory {
+        id: &'static str,
+        addr: u32,
+        len: u32,
+    },
+    Illegal,
+}
```
FanShupei, line 505: 1. OutOfMemory means the system does not have enough memory to perform malloc. It's not the right term here.
2. When is 'Illegal' intended to use?

pokedex/simulator/pokedex/src/simulator.rs
Avimitin: In my opinion methods naming is better to start with verb, and each read/write to a MMIO device is actually require a response from them.

FanShupei: > In my opinion methods naming is better to start with verb

If so, may use `do_bus_read`

> and each read/write to a MMIO device is actually requiring a response from them.

The response is exactly the return value. The function call perform both request and response, only name it 'resp' is misleading.

FanShupei (approved):

PR #1159, opened 2025-07-22
[pokedex-model] RVV: implement integer arithmetic instructions

FanShupei:


FanShupei:
To reduce massive replications, this PR introduces generating from templates. Now the ASL implementation for instructions could be generated from jinja templates.

Key/values are in toml files for each instruction. The jinja template is fixed to `inst.j2`. It dispatches to the real template depending on the `template` key. Using a fixed template simplifies the build system. We use `minijinja-cli` as the template engine. See `generate_inst_from_template.sh` for how to invoke it.

TODO:
- Currently auto-generated files are committed. They shall be removed in future build system refactor.
- `generate_inst_from_template.sh` could be replaced by auto-discovering toml files.

PR #1182, opened 2025-10-23
[pokedex] add CSR difftest and fix CSR for rv32imafc at M mode

Avimitin:
- **[pokedex] rewrite model to log CSR value after write**
- **[pokedex-difftest] add CSR comparison**
- **[pokedex] rewrite CSR handling in instruction in a more clean way**
- **[pokedex-tests] disable Spike debug CSR usage**

### DIFF
--- a/pokedex/simulator/batchrun/batchrun.py
+++ b/pokedex/simulator/batchrun/batchrun.py
@@ -111,7 +111,7 @@ if __name__ == "__main__":
         config = cast(BatchrunConfig, json.loads(raw_config))
 
     ignore_match = IgnoreTestExt(config["ignore_tests"])
-    context = []
+    failed_log = []
 
     for elf in glob.glob(config["elf_path_glob"], recursive=True):
         elf_path = Path(elf)
@@ -130,12 +130,15 @@ if __name__ == "__main__":
                 else:
                     print(f"\n** {RED}FAIL{RESET}   {elf_path.name}\n")
                     print(result["context"])
-                    context.append(elf_path)
+                    failed_log.append(str(spike_log_path))
+                    failed_log.append(str(pokedex_log_path))
+                    failed_log.append(str(difftest_result))
+                    failed_log.append(str(elf_path))
 
-    if len(context) != 0:
+    if len(failed_log) != 0:
         print("** batch run fail on some tests **")
         with open("batch-run-result.json", "w") as file:
-            file.write(json.dumps(context))
+            file.write(json.dumps(failed_log))
         exit(1)
     else:
         print("** all tests pass **")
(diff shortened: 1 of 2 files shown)


pokedex/model/handwritten/states.asl
```diff
@@ -140,11 +159,46 @@ begin
...
+
+func _internal_set_mstatus_sd()
+begin
+  // TODO: add VS, XS when supported
+  if _MSTATUS_FS == '11' || _MSTATUS_VS == '11' then
+    MSTATUS_SD = '1';
+  elsif _MSTATUS_FS != '11' && _MSTATUS_VS != '11' then
+    MSTATUS_SD = '0';
+  end
+end
```
FanShupei: I prefer to compute mstatus.SD on the fly at reading, instead of syncing it in every write.

pokedex/model/template/csr/write/fcsr_003.asl
```diff
@@ -1,4 +1,15 @@
-FFLAGS = value[4:0];
-FRM = value[7:5];
+var mr = ReadCSR(MSTATUS_IDX);
+assert mr.is_ok;
+mr.value[MSTATUS_FS_HI:MSTATUS_FS_LO] = '11'; // set dirty
+let mw = WriteCSR(MSTATUS_IDX, mr.value);
+assert mw.is_ok;
```
FanShupei: This could be abstracted into a function, e.g. makeFsDirty. Every F instructions will contain such logics.

pokedex/simulator/difftest/src/pokedex.rs
```diff
@@ -66,3 +61,60 @@ impl Display for InsnCommit {
...
+                }
+                Csr { idx, name, value } => {
+                    if state.write_csr(name, *idx, *value).is_some() {
+                        if has_frf_write {
+                            check_ty.csr_mask = crate::replay::CsrCheckType::FpCsrOnly;
+                        } else {
+                            check_ty.csr_mask = crate::replay::CsrCheckType::AllCsr;
+                        }
+                    }
+                }
```
FanShupei: The logic to determine csr_mask looks too confusing. Consider adding a new `has_csr_write` local variable, then computing csr_mask from has_frf_write and has_csr_write at the very end?

pokedex/simulator/difftest/src/replay.rs
```diff
@@ -0,0 +1,213 @@
...
+{
+    commit_cursor: &'a mut Peekable<Iter<'b, T>>,
+    state: CpuState,
+}
+
+pub trait IsCpuState {
+    fn get_pc(&self) -> u32;
+    fn has_reset(&self) -> Option<u32>;
+    fn write_cpu_state(&self, state: &mut CpuState) -> StateCheckType;
+}
```
FanShupei: Maybe it should be called "IsCommit". I see it's impl-ed in Commit and InsnCommit, not CpuState.

PR #1193, opened 2025-11-13
[Pokedex] Define explicit infterface between ASL model and simulator

FanShupei:
The interface is defined in `pokedex/simulator/pokedex/include/pokedex_vtable.h`, in the first HUGE commit

Now the ASL model is compiled to self-contained dynamic lib, with libASL / libsoftfloat bundled.
The dylib has only one export symbol `EXPORT_ASL_MODEL_get_pokedex_vtable`.

One benefit for dylib is that it avoids potential symbol collisions. asl2c does not do name mangling and global variables are everywhere. Now they are all encapsulated.

In the second HUGE commit. The simulator is rewritten to adapt to the new interface, with a more clean architecture. It's better to review the code directly rather than the diff. 

TODO:
- LR/SC support is half backed. Plan to improve it later.
- Temporally remove memory access in pokedex json log.
  I'm exploring how to coalesce memory access logs for RVV load/store.
- The simulator actually supports bundled (statically-linked) ASL model,
  but ASL model side is not prepared yet. Do we actually need this feature in the future?

### DIFF
--- a/pokedex/tests/runner.nix
+++ b/pokedex/tests/runner.nix
@@ -2,6 +2,7 @@
   runCommand,
   spike,
   dtc,
+  model,
   simulator,
   jq,
 }:
@@ -14,6 +15,10 @@ runCommand "difftest-${caseInfo.caseName}"
       dtc
       jq
     ];
+
+    env = {
+      POKEDEX_MODEL_DYLIB = "${model}/lib/libpokedex_model.so";
+    };
   }
   ''
     runSpike() {


pokedex/model/package.nix
```diff
@@ -22,6 +51,14 @@ stdenv.mkDerivation {
...
     RISCV_OPCODES_SRC = "${riscv-opcodes-src}";
+
+    SOFTFLOAT_RISCV_INCLUDE = "${softfloat-riscv}/include";
+
+    SOFTFLOAT_RISCV_LIB = "${softfloat-riscv}/lib/libsoftfloat.a";
+
+    # Do not let model depend on other parts of pokedex in nix build,
+    # therefore directly pull the include directory.
+    POKEDEX_INCLUDE = "${../simulator/pokedex/include}";
   };
```
Avimitin: It will be better to add a passthrough here:

```nix
passthru = { inherit softfloat-riscv; };
```

Avimitin (changes requested):

PR #1202, opened 2025-11-21
[poekdex] Redesign trace prototols

FanShupei:
This PR redesigns the interface. Changes can be summarized as follows:
- Add various accessors to inspect arch states
- For tracing state writes, it will only records the index. The value could be retrieved through accessors after step.
- State writes now are recorded to trace buffer, instead of through callbacks. It simplifies the C/Rust interface.

These accessors are indented used by "debugger". They always return the corresponding arch states  regardless of current privilege mode. They never mutate arch states nor have any other side effects. For read CSRs, it's implemented by GetRaw_XXX functions.

Other changes in ASL model and simulator should be straightforward, though massive.

As the branch name suggests, I'm working on adding GDB remote protocol support, to ease the software development.
It will be done in following PRs. Adding accessors is a prerequisite for it.

### DIFF
--- a/pokedex/simulator/src/difftest/mod.rs
+++ b/pokedex/simulator/src/difftest/mod.rs
@@ -90,8 +90,8 @@ fn run_diff(
 
         match (&status1, &status2) {
             (Status::Running(dr1), Status::Running(dr2)) => {
-                let state1 = source2.state();
-                let state2 = source1.state();
+                let state1 = source1.state();
+                let state2 = source2.state();
 
                 let combined_dr = DiffRecord::combine(dr1, dr2);
 
--- a/pokedex/simulator/src/difftest/replay.rs
+++ b/pokedex/simulator/src/difftest/replay.rs
@@ -169,7 +169,10 @@ pub fn pretty_print_diff(
 ) -> std::fmt::Result {
     if gold.pc != dut.pc {
         writeln!(f, "pc        : {:010x} <-> {:010x}", gold.pc, dut.pc)?;
+    } else {
+        writeln!(f, "pc        : {:010x}", gold.pc)?;
     }
+
     // compare GPR
     for i in 0..32 {
         let (goldv, dutv) = (gold.gpr[i], dut.gpr[i]);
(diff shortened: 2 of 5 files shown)


pokedex/model/csrc/pokedex_interface.c
```diff
@@ -207,34 +234,126 @@ static void* pokedex_create(
...
+static const struct pokedex_export export = {
+    .abi_version = POKEDEX_ABI_VERSION,
     .create = pokedex_create,
     .destroy = pokedex_destroy,
 
+    .get_description = pokedex_get_description,
+
     .reset = pokedex_reset,
-    .step = pokedex_step,
+    .step = pokedex_step_trace,
```
Avimitin: Do we need to keep `step` or `step` is for future design?

FanShupei: Yes, actually `step` is the primary method. When we are not doing difftest, `step` is enough, e.g. use pokedex only to run testcases, or act as an gdb stub when doing interactive debugging.

`step_trace` is an optimization to speed up difftest. The caller could simply call `step` then read everything through accessors.

In our current implementation, since ASL is too slow, tracing adds little overhead, so no bother to write a dedicated "step without trace". But I'd like make the distinction clear in concept.

pokedex/model/csrc/pokedex_interface.c
```diff
@@ -207,34 +234,126 @@ static void* pokedex_create(
...
+    .vlen = 256,
+};
+
+const struct pokedex_model_description* pokedex_get_description(void* _model) {
+    (void)(_model);
+
+    return &model_desc;
+}
+
+static const struct pokedex_export export = {
```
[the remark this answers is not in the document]
FanShupei: Thanks for your suggestion. "pokedex_export" is renamed to "pokedex_model_export", I'll keep it prefixed with pokedex since it's C's way to do namespacing. Methods are renamed to "model_get_xreg", etc.

The export entry is renamed to "EXPORT_pokedex_get_model_export". The interface is quite generic, no need to emphasize ASL here.

FanShupei:
All comments are addressed. Once the review is done, I'll do a rebase to squash all fixups.

Avimitin (approved):

Avimitin (approved):

PR #1205, opened 2025-11-25
[pokedex] Initial GDB support

FanShupei:
# Initial GDB Support

To support the feature, I have to do a major code refactor first. I float "model" and "bus" implementation to the crate root. No functional change is made to existing features.

## Feature Status

Current implementation is very INCOMPLETE. Notably, it does not report ~vector registers~ and privilege CSRs to gdb. Fixing the problem requires providing the correct target xml, which is non-trivial. (To support different configs, constructing the target xml dynamically seems inevitable.)

UPDATE: support vector registers now, but due to constraints of gdbstub, VLEN is fixed at compile time.

Besides this, since gdb remote protocol is very fragile and not every detail is documented, the implementation is expected to have bugs here and there. But at least it roughly works now.

## Usage

```
POKEDEX_LOG_LEVEL=trace pokedex debug <elf_path>
```

It will listen to `localhost:1234` for gdb connection. The port could be changed using `-p` option.

With log level set to trace, gdbstub crate will log all response/reply messages.

Open another terminal run `gdb`, then run `target remote :1234` to connect to pokedex.


pokedex/simulator/src/gdb/rv32f.xml
Avimitin: Do we need this file? since gdb/arch.rs is already a build script

Avimitin (dismissed):

Avimitin (approved):

PR #1208, opened 2025-12-04
Clean up pokedex

Avimitin:
- **[pokedex] drop rvopcode-cli**
- **[pokedex] drop useless build file**
- **[pokedex] drop useless makefile**
- **[pokedex] pass config.toml to datagen.py**
- **[nix] add black as Python formatter**


pokedex/model/scripts/ninja_syntax.py
FanShupei: I notice all python files are formatted, is that done munually or by tools? In either case, I suggest `ninja_syntax.py` shall be excluded, since it's literally vendored from upstream.

FanShupei (approved):

PR #1211, opened 2025-12-09
[pokedex] refactor CSR decoder with document comment metadata

Avimitin:
- **[pokedex] refactor CSR to hold metadata**
- **[pokedex] migrate CSR decoder from JSON to metadata**
- **[pokedex] drop csr.json**


pokedex/model/csr/mro_f11_mvendorid.asl
```diff
@@ -1,3 +1,18 @@
+//! ---
+//! csr: "mvendorid"
+//! bin: "111100010001"
```
FanShupei: "bin" is redundant with id, please remove it. If it's needed, it could be computed at datagen.

Actually, we could totally replace it by`{{ "{:012b}".format(csr_id) }}`. To make it work, we should use use the newer minjinja (we have 2.10 in nix, it does not work, the newest is 2.13) with  `--py-compat` CLI option.

PR #1213, opened 2025-12-19
[pokedex] migrate tests to meson

Avimitin:
## Current Issues

- Makefile is not maintainable
  * Tons of copy paste program searching logic and options
  * Unreadable string manipulation
  * Unreadable build output
  * Unmaintainable build graph and dependencies
  * Project get their own artifacts install logic instead of keeping build output in one place
- Nix as test framework is not ideal
  * Each test require a call to bash + bwrap + chroot + tons of cp......etc
  * Unmaintainable tests discovery logic
  * Unreadable tests output
  * Hard to do rapid code evolution
  * Hard to remember all the commands (Imaging getting a diff test env require `nix develop .#pokedex.full.tests.smoke-tests.passthru.diff.addi`)

## Solution

- Meson as the main tests build system
  * It has a top level build script entry and all common options can be shared downwards
  * It has Python syntax
  * It has pretty output combined with Ninja
  * It has declarative-like syntax and we don't need to manage build graph and dependencies
  * It support `subdir()` to include a test suites own build logic and grouping them together
- Meson as test framework
  * It run program with just `execve`
  * Test discovery nears build logic, easy to find and customize.
  * It has pretty output
  * We can now run tests with `nix develop .#pokedex.zve32x.tests.env` for environment and a simple meson setup with meson test to run all the tests

### DIFF
--- a/pokedex/tests/README.md
+++ b/pokedex/tests/README.md
@@ -1,0 +1,261 @@
+# Pokedex Test Suite Technical Manual
+
+This manual documents the architecture, usage, and configuration of the Pokedex
+test suite. This project validates the Pokedex RISC-V simulator by running
+differential tests against the Spike simulator.
+
+## 1. Quick Start (Nix)
+
+The recommended way to run tests is using **Nix**. It handles toolchains,
+dependencies, and configuration automatically.
+
+### Running Tests
+To run the full test suite via Nix:
+
+```bash
+nix build '.#pokedex.zve32x.tests.run' -L
+```
+
+### Development Environment
+For interactive development (compiling individual tests, debugging):
+
+1.  **Enter the Shell**:
+    ```bash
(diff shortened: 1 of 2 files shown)


Avimitin:
In short, riscv-vector-tests do use compile-stub and the `_start` function is in `.text.init` section.

The compile-stub include directory vendored the riscv-tests-env. Generated code from riscv-vector-tests will include headers from riscv-tests-env. Each generated case has a `RVTEST_CODE_BEGIN` macro at the beginning of the file. After macro expand, each case will eventually include a code snippet for boot up, which are placed in `.text.init` section.

image

There are two entry point in current tests. The `main.S` is to simplify writing smoke case. For convention, the `riscv_test.h` will be used for riscv-tests test suite and riscv-vector-tests test suite. I will add a document for this.

FanShupei (dismissed):

PR #1216, opened 2026-01-27
eval: add new softmax case

Avimitin:


FanShupei:
We should separate dlen=4096 changes from real bug fixes and configuration improvements.

DLEN=4096 AXI is a temporary hack. Everything only for enabling dlen=4096 for should live in a dedicated branch, which probably will never be merged into master.

Avimitin:
DLEN 4096 modification moved to #1217

PR #1218, opened 2026-02-05
Fix idea

Avimitin:


Avimitin:
Closed in favor of zed editor with metals.

